PR #2, opened 2018-03-31
Add function to get listens and user RDD from uncompressed LB dump

ferbncode:
Added function for loading data into spark and creating user and listens RDD from uncompressed LB dump.

### DIFF
--- a/load_data.py
+++ b/load_data.py
@@ -50,6 +50,7 @@ def load_listenbrainz_dump(directory, sc):
                     bytes_read += len(line)
                     listen = json.loads(line)
                     listen["user_name"] = user['user_name']
+                    listen = (listen["user_name"], listen["listened_at"], listen["recording_msid"], listen["track_metadata"]["track_name"], json.dumps(listen["track_metadata"]))
                     listens.append(listen)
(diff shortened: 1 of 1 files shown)


load_data.py
```diff
@@ -1,28 +1,132 @@
...
+                while bytes_read < user['size']:
+                    line = f.readline()
+                    bytes_read += len(line)
+                    listen = json.loads(line)
+                    listen["user_name"] = user['user_name']
+                    listens.append(listen)
+                    if len(listens) > IMPORT_CHUNK_SIZE:
+                        listens_rdd = listens_rdd.union(sc.parallelize(listens))
+                        listens = []
+                    if len(listens) > 0:
```
mineo: Shouldn't these two lines be indented one level less, so they only add the listens that are left after a users listens have been read completely, but the last batch didn't include `IMPORT_CHUNK_SIZE` listens? At the moment, it looks like after reading one line, all the lines that have been read for the current batch will be added to `listens_rdd`, over and over again (`RDD.union` does not eliminate duplicates).

mayhem: Looks to be a problem yes. Fortunately this code is already out of date. :)

PR #9, opened 2018-12-10
Add spark-writer to write new listens to HDFS

paramsingh:
There is still a lot of stuff to think about, but this is a good
first cut in my opinion.

There is still some stuff to work out.

- [ ] Discuss directory structure and file sizes while storing in HDFS
- [ ] Figure out how to manage configurations in the cluster

(and other things that'll come up)


listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class SparkWriter:
+
+    def write_message_to_hdfs(self, listens):
+        current_app.logger.error(json.dumps(listens, indent=4))
+        data = self._convert_listens_to_string(listens)
+        path = self._create_path(self.current_file)
+        hdfs_connection.client.write(path, data=data, append=True)
+        if self._get_size(self.current_file) > HDFS_FILE_SIZE_LIMIT:
```
mayhem: This seems like a bad idea -- do a write and THEN check to see if the file size is exceeded? That check should be done BEFORE the write, no?

listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,95 @@
...
+    def callback(self, channel, method, properties, body):
+        listens = json.loads(body.decode('utf-8'))
+        self.write_message_to_hdfs(listens)
+        channel.basic_ack(delivery_tag=method.delivery_tag)
+
+    def run(self):
+        with create_app().app_context():
+            while True:
+                hdfs_connection.init_hdfs(current_app.config['HDFS_NAMENODE_URI'])
+                self.current_file = self._new_file()
```
mayhem: Does this mean that everytime we start this container it starts a new file?

listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,95 @@
...
+                rabbitmq = init_rabbitmq(
+                    username=current_app.config['RABBITMQ_USERNAME'],
+                    password=current_app.config['RABBITMQ_PASSWORD'],
+                    host=current_app.config['RABBITMQ_HOST'],
+                    port=current_app.config['RABBITMQ_PORT'],
+                    vhost=current_app.config['RABBITMQ_VHOST'],
+                    log=current_app.logger.critical,
+                )
+                channel = rabbitmq.channel()
+                channel.exchange_declare(exchange=current_app.config['INCOMING_EXCHANGE'], exchange_type='fanout')
```
mayhem: I see no place where these are defined...

listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,95 @@
...
+
+                current_app.logger.info('Started spark-writer...')
+                try:
+                    channel.start_consuming()
+                except pika.exceptions.ConnectionClosed:
+                    current_app.logger.error('connection to rabbitmq closed.', exc_info=True)
+                    continue
+                rabbitmq.close()
+
+    def _create_path(self, filename):
```
mayhem: I guess this goes with trying to determine how to store files -- with this approach we have no way to partition the data in any sort of way. If we wanted to process something based on the last x weeks, we still need to process all the data. Let's discuss in IRC.

listenbrainz_spark/spark_writer/spark_writer.py
paramsingh: hmm, I was going with the method we used for the dumps where we stopped as soon as the file size exceeds 1 GB.

listenbrainz_spark/spark_writer/spark_writer.py
paramsingh: hmm, bug. should have thought of that. I'll change it.

listenbrainz_spark/spark_writer/spark_writer.py
mayhem: Ah, I see. HDFS_FILE_SIZE_LIMIT is our limit, not hadoops?

listenbrainz_spark/spark_writer/spark_writer.py
paramsingh: Defined in the config file. https://github.com/metabrainz/listenbrainz-recommendation-playground/pull/9/files#diff-0465562b65e73c30c981c975bcf55b61R15

paramsingh:
Ok, i've refactored this PR to write to already existing dataframes in hdfs in the format in which data was imported. it works locally for me, gonna test on the cluster soon.

listenbrainz_spark/utils.py
```diff
@@ -8,3 +16,34 @@ def create_path(path):
     except OSError as e:
         if e.errno != errno.EEXIST:
             raise
+
+
+def create_app():
+    app = CustomFlask(import_name=__name__)
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'config.py')
+    for i in range(APP_CREATION_RETRY_COUNT):
```
mayhem: If this loop is exited it will likely die with a stacktrace later, which isn't so nice. And else clause on the for loop could take care of printing a message and.. exiting?

listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,94 @@
...
+        self.write_message_to_hdfs(listens)
+        channel.basic_ack(delivery_tag=method.delivery_tag)
+
+    def run(self):
+        app = create_app()
+        if not app:
+            sys.exit(-1)
+        with app.app_context():
+            while True:
+                hdfs_connection.init_hdfs(current_app.config['HDFS_HTTP_URI'])
```
mayhem: Could this fail? if so, we need to catch an exception.

listenbrainz_spark/spark_writer/spark_writer.py
```diff
@@ -0,0 +1,94 @@
...
+                try:
+                    channel.start_consuming()
+                except pika.exceptions.ConnectionClosed:
+                    current_app.logger.error('connection to rabbitmq closed.', exc_info=True)
+                    continue
+                rabbitmq.close()
+
+def main(app_name):
+    listenbrainz_spark.init_spark_session(app_name)
+    SparkWriter().run()
```
mayhem: I think we may want to add a catch-all exception here and log to sentry, then restart.

listenbrainz_spark/utils.py
```diff
@@ -8,3 +16,39 @@ def create_path(path):
...
+
+def create_app():
+    app = CustomFlask(import_name=__name__)
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'config.py')
+    for _ in range(APP_CREATION_RETRY_COUNT):
+        if os.path.exists(config_file):
+            break
+        sleep(1)
+    else:
+        print("Config file does not exist!")
```
mayhem: use logging, not print.

paramsingh:
Closing this for now after the incremental data dumps discussion.

PR #11, opened 2019-01-16
Job runner

mayhem:
Create a script that allows easy running of spark jobs.


run.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+if [ "$#" -ne 1 ]; then
```
paramsingh: this should be `-lt 1`, so that large commands such as `./run.sh spark-submit blah blah blah` work too.

PR #15, opened 2019-01-28
Delete temporary directories from playground and hdfs

vansika:
**PROBLEM:**- Temporary directories created to process 'JSON' files have neither been deleted in playground container nor in hdfs. 

**SOLUTION:-** issue commands to delete the directories after data processing is done.

### DIFF
--- a/listenbrainz_spark/data/import_dump.py
+++ b/listenbrainz_spark/data/import_dump.py
@@ -88,6 +88,8 @@ def copy_to_hdfs(archive, threads=8):
                 total_time += time_taken
                 average_time = total_time / file_count
                 print("Total time: %.2f, average time: %.2f" % (total_time, average_time))
+    hdfs_connection.client.delete(tmp_dump_dir, recursive=True)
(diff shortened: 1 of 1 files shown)


listenbrainz_spark/data/import_dump.py
```diff
@@ -88,6 +88,8 @@ def copy_to_hdfs(archive, threads=8):
                 total_time += time_taken
                 average_time = total_time / file_count
                 print("Total time: %.2f, average time: %.2f" % (total_time, average_time))
+    hdfs_connection.client.delete(tmp_dump_dir, recursive=True)
+    shutil.rmtree(tmp_dump_dir)
```
mayhem: Why is this here? The local filesystem should not be affected by copying the files to HDFS, since files are output to stdout. Or am I missing something here?

paramsingh: we create a tmp dir using `tempfile.mkdtemp` which creates a directory locally, need to remove that as well.

PR #16, opened 2019-01-31
Queries to load DataFrames specific to user 

vansika:
Queries to get artists, recordings and tracks of a specified user in sorted order for the previous month.


listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
+import sys
+import listenbrainz_spark
+import time
+from datetime import datetime  
+
+
+def top_artist(user_name):
+	playcounts_df = listenbrainz_spark.sql_context.sql("""
+            SELECT artist_name, count(track_name) as cnt
```
paramsingh: count(artist_name)

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+              FROM listen
+             WHERE user_name = '%s'
+          GROUP BY artist_name
+          ORDER BY cnt DESC
+             LIMIT 20
+        """ % user_name)
+	playcounts_df.show()
```
paramsingh: The results should be returned as a dictionary.

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+
+def top_recording(user_name):
```
paramsingh: `get_top_recordings`

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+             LIMIT 20
+        """ % user_name)
+	playcounts_df.show()
+
+def top_release(user_name):
+	playcounts_df = listenbrainz_spark.sql_context.sql("""
+            SELECT release_name, count(release_name) as cnt
+              FROM listen
+             WHERE user_name = '%s'
+          GROUP BY release_name
```
paramsingh: this should be grouped by release_msid

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+             WHERE user_name = '%s'
+          GROUP BY track_name
+          ORDER BY cnt DESC
+             LIMIT 20
+        """ % user_name)
+	playcounts_df.show()
+
+def top_release(user_name):
+	playcounts_df = listenbrainz_spark.sql_context.sql("""
+            SELECT release_name, count(release_name) as cnt
```
paramsingh: we should get as much useful data in here as possible. See queries here (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/stats/user.py)

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+		print ("No Listens for last month")
+		sys.exit(-1)
+	df.printSchema()
+	print(df.columns)
+	print(df.count())
+	df.registerTempTable('listen')
+	print("Running Query...")
+	query_t0 = time.time()
+	print("DataFrame loaded in %.2f s" % (time.time() - query_t0))
+	run_queries(user_name)
```
paramsingh: would prefer if you printed times to run the queries here

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,62 @@
...
+	playcounts_df.show()
+
+def top_release(user_name):
+	playcounts_df = listenbrainz_spark.sql_context.sql("""
+            SELECT release_name, count(release_name) as cnt
+              FROM listen
+             WHERE user_name = '%s'
+          GROUP BY release_name
+          ORDER BY cnt DESC
+             LIMIT 20
```
paramsingh: this limit can be taken from the config.

vansika:
@mayhem I think I will need list of all registered users so as to update db for users who 0 listens for previous month. Should be exported with listen data.

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns:
+        stats (list): list of dicts where each dict is a message/packet
+                    for RabbitMQ containing user stats 
+    """
+    t0 = time.time()
+    listenbrainz_spark.init_spark_session(app_name)
+    prev_month = month - 1 if month > 1 else 12 
+    curr_year = year if prev_month < 12 else year - 1 
+    try:
+        df = listenbrainz_spark.sql_context.read.parquet('hdfs://hadoop-master:9000/data/listenbrainz/{}/{}.parquet'.format(curr_year, prev_month))
```
mayhem: These values should be taken from config.py

listenbrainz_spark/RabbitMQ/rabbitmq_connection.py
```diff
@@ -0,0 +1,61 @@
+import pika
+import logging
+from listenbrainz_spark import config
+import time
+import json
+import sys
+
+class SparkReader():
```
mayhem: A class called SparkReader() in the rabbitmq_connection is weird to me. It is specifically used to write queried user stats to RabbitMQ... Should it be called StatsWriter() or RabbitMQWriter()? And should the module get a better name?

listenbrainz_spark/RabbitMQ/rabbitmq_connection.py
```diff
@@ -0,0 +1,61 @@
...
+    def start(self, data):
+        """Publish data to RabbitMQ
+        """
+        if "RABBITMQ_HOST" not in config.RABBITMQ:
+            logging.critical("RabbitMQ service not defined. Sleeping {0} seconds and exiting.".format(self.ERROR_RETRY_DELAY))
+            time.sleep(self.ERROR_RETRY_DELAY)
+            sys.exit(-1)
+
+        try:
+            self.connect_to_rabbitmq()
```
mayhem: The next few lines get called for each time we write data to RabbitMQ. Each time we re-setup the connection to RabbitMQ, which takes time. I think it makes more sense to check to see if we have already established a connection (maybe test for unique_ch being not None?) and then try and use the connection right away. If the write to RabbitMQ fails, then re-do the connection.

So, I think this would best be split into two function:
- start() -- setup the connections, possibly re-setting up any timedout connections.
- write() -- write data to the connection, but if the write fails, call start() again?

listenbrainz_spark/RabbitMQ/rabbitmq_connection.py
```diff
@@ -0,0 +1,61 @@
...
+                    credentials=credentials,
+                )
+                self.connection =  pika.BlockingConnection(connection_parameters)
+                break
+            except Exception as err:
+                error_message = "Cannot connect to RabbitMQ: {error}, retrying in {delay} seconds."
+                print(error_message.format(error=str(err), delay=self.ERROR_RETRY_DELAY))
+                time.sleep(self.ERROR_RETRY_DELAY)
+
+    def start(self, data):
```
mayhem: this should be called write(), rather than setup(), no?

listenbrainz_spark/RabbitMQ/rabbitmq_connection.py
paramsingh: StatsWriter and the module `stats_writer/stats_writer.py` would look good to me.

mayhem: Agreed!

vansika: I will say StatsWriter. What about you? 
Would you like to suggest any name for the module?

mayhem: As param suggested, would be great!

listenbrainz_spark/RabbitMQ/rabbitmq_connection.py
vansika: Yes. Sounds good:)

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,193 @@
+import sys
+import listenbrainz_spark
+import time
+from datetime import datetime  
+from listenbrainz_spark.stats_writer.stats_writer import StatsWriter
+from listenbrainz_spark import config
+
+LIMIT = 100
+month = datetime.now().month
```
mayhem: Why are these two here? They are only used in main, so they should be set there.

listenbrainz_spark/stats_writer/stats_writer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def start(self, data):
+        """Establishes connection to RabbitMQ if not connected
+        """
+        if not config.RABBITMQ_HOST:
+            logging.critical("RabbitMQ service not defined. Sleeping {0} seconds and exiting.".format(self.ERROR_RETRY_DELAY))
+            time.sleep(self.ERROR_RETRY_DELAY)
+            sys.exit(-1)
+
+        if self.unique_ch:
```
mayhem: There exists a case where we might have opened a connection and have a unique_ch object -- however, that object might be stale and the connection to RabbitMQ has been dropped. This means the subsequent write() will fail and right now that error will fail quietly. 

I think rewriting this function to:

0. In an infinite loop:
1. Check to see if we have a connection by checking unique_ch, if not open one by calling connect_to_rabbitmq
2. Attempt the write. If it works, exit the loop.
3. If the write fails, clean up the existing connection (if we can/need) and set unique_ch to None. Then repeat the loop.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 click==6.2
 hetznercloud==1.1.1
 hdfs == 2.1.0
+pika == 0.12.0
```
paramsingh: the latest version is 0.13.0, so we should probably use that here.

listenbrainz_spark/config.py.sample
```diff
@@ -1,3 +1,12 @@
 HDFS_HTTP_URI = 'http://hadoop-master:9870' # the URI of the http webclient for HDFS
 
 HDFS_CLUSTER_URI = 'hdfs://hadoop-master:9000' # the URI to be used with Spark
+
+RABBITMQ_HOST: "rabbitmq"
```
paramsingh: these should be `=` not `:`

vansika:
@paramsingh  The filename of JSON files which contain user data and have been exported to LB-playground should be padded.
1.json should be 01.json. Please do the necessary changes so that I can work on sending timestamps with stats.

paramsingh:
this has been replaced by #21

PR #21, opened 2019-02-25
calculate stats for all users at once

vansika:
Calculates stats by running a single query for all users.


listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+    print("time taken by collect call: %.2f" % (time.time() - t))
+    for row in rows:
+        artists[row.user_name].append({
+            'artist_name': row.artist_name,
+            'artist_msid': row.artist_msid,
+            'artist_mbids': row.artist_mbids,
+            'listen_count': row.cnt,
+        })
+    query_t0 = time.time()
+    print("Query to calculate artist stats processed in %.2f s" % (query_t0 - t0))
```
paramsingh: Why aren't we returning anything here?

paramsingh: Or in any of the other get functions

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def get_users(table):
+    """ DataFrame is registered as a temporary table
```
paramsingh: Is this function still needed?

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+                  application amongst other applications
+    """
+    t0 = time.time()
+    listenbrainz_spark.init_spark_session(app_name)
+    t = datetime.utcnow().replace(day=1)
+    date = t + relativedelta(months=-1)
+    try:
+        df = listenbrainz_spark.sql_context.read.parquet('{}/data/listenbrainz/{}/{}.parquet'.format(config.HDFS_CLUSTER_URI, date.year, date.month))
+        print("Loading dataframe...")
+    except:
```
mayhem: A blanket exception catch is not a good idea in 99% of cases -- it often hides other exceptions, as it does in this case. If the script cannot read the parqet file, it will happily tell us that there are no new listens. Whoops.

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+    users = get_users(table)
+    print("Number of users: %d" % len(users))
+    data = defaultdict(dict)
+    artist_data = get_artists(table)
+    for user_name, artist_stats in artist_data.items():
+        data[user_name]['artists'] = {
+            'artist_stats': artist_stats,
+            'artist_count': len(artist_stats),
+        }
+    recording_data = get_recordings(table)
```
mayhem: Check for exceptions.

mayhem: ... here and all the other places in this function ...

vansika: I am not able to think of any exceptions here. If the parquet does not exist or there is any problem in reading it we would catch it earlier in main, otherwise we would have all the stats. Am I missing out anything?

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+    return users
+
+def main(app_name):
+    """
+    Args:
+        app_name: Application name to uniquely identify the submitted
+                  application amongst other applications
+    """
+    t0 = time.time()
+    listenbrainz_spark.init_spark_session(app_name)
```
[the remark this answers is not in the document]
vansika: There may be diff kinds of exceptions associated with it. I will prefer to catch them and print, the error message can be used to diagnose the problem. Or is there anything else you would like to suggest?

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,190 @@
...
+    release_data = get_releases(table)
+    for user_name, release_stats in release_data.items():
+        data[user_name]['releases'] = release_stats
+    rabbbitmq_conn_obj = StatsWriter()
+    yearmonth = datetime.strftime(date, '%Y-%m')
+    for user_name, metadata in data.items():
+        data[user_name]['yearmonth'] = yearmonth
+        rabbitmq_data = {}
+        rabbitmq_data[user_name] = metadata
+        rabbbitmq_conn_obj.start(rabbitmq_data)
```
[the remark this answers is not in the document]
vansika: I would but I am not sure if I understand the purpose.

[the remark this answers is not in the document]
vansika: I would but I am not sure if I understand the purpose.

listenbrainz_spark/stats/user.py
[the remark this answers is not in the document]
mayhem: Yes, you need to find each and every exception that it is likely to throw and you need to catch each one. Blanket exceptions like this are bad and can only be used in very special situations -- this is not one of them.

vansika:
@mayhem Flow of data is as follows:
Request for stats from LB  server via rabbitmq -> user.py -> stats_writer for pushing stats in queue -> stats delivered from LB playground to sever.
If there is a problem with stats_writer we should catch it and inform the script requesting stats from LB server. Is that you want to say? If yes, then the warning/error should be sent to server using rabbitmq?

listenbrainz_spark/stats/user.py
```diff
@@ -17,8 +17,12 @@ def get_artists(table):
         table: name of the temporary table
 
     Returns:
-        artists: list containing information of artists 
-                grouped by users
+        artists: A dict of dicts containing artists information of every user
```
paramsingh: This description is good, but the function still needs a line in the beginning of the docstring explaining what it does.

we follow the format mentioned here: https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings

Consistency is important as it makes things easier to understand. :)

listenbrainz_spark/stats/user.py
```diff
@@ -131,42 +148,55 @@ def main(app_name):
...
+
     table = 'listens_{}'.format(datetime.strftime(date, '%Y_%m'))
     print(table)
     df.registerTempTable(table)
     print("Running Query...")
     query_t0 = time.time()
     print("DataFrame loaded in %.2f s" % (query_t0 - t0))
+
     data = defaultdict(dict)
+    """
```
paramsingh: Comments and docstrings are different in python. 

https://stackoverflow.com/a/19076001

This should probably be a comment using hashes (`#`) and not docstrings.

listenbrainz_spark/stats/user.py
vansika: have removed the function.

listenbrainz_spark/stats/user.py
```diff
@@ -175,5 +205,10 @@ def main(app_name):
...
         }
-        rabbbitmq_conn_obj.start(rabbitmq_data)
-    sys.exit(0)
+        try:
+            rabbbitmq_conn_obj.start(rabbitmq_data)
+            print("Statistics of %s pushed to rabbitmq" % (user_name))
+        except Exception as err:
+            logging.error("Cannot publish statistics of %s to rabbitmq channel: %s / %s." % (user_name, type(err).__name__, str(err)), exc_info=True)
+            continue
+
```
paramsingh: extra newlines

listenbrainz_spark/stats_writer/stats_writer.py
```diff
@@ -30,7 +30,7 @@ def connect_to_rabbitmq(self):
                 )
                 self.connection =  pika.BlockingConnection(connection_parameters)
                 break
-            except Exception as err:
+            except:
```
paramsingh: this will error out because `err` is used in the except block. Why did you change this?

paramsingh: It is more pythonic to be specific, so I'd prefer to go back to except Exception as err.

listenbrainz_spark/stats/user.py
```diff
@@ -131,42 +148,55 @@ def main(app_name):
     Args:
         app_name: Application name to uniquely identify the submitted
                   application amongst other applications
+
+        The script will be run to calculate stats for previous month.
```
paramsingh: The docstring is for the main function so referring to it as the script does not make sense. Also, this sentence should be at the top of the docstring.

listenbrainz_spark/stats/user.py
vansika: yeah. right. Thanks for the link :)

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,214 @@
...
+        df = listenbrainz_spark.sql_context.read.parquet('{}/data/listenbrainz/{}/{}.parquet'.format(config.HDFS_CLUSTER_URI, date.year, date.month))
+        print("Loading dataframe...")
+    except Exception as err:
+        logging.error("Cannot read files from HDFS: %s / %s. Aborting." % (type(err).__name__, str(err)))
+        sys.exit(-1)
+    df.printSchema()
+
+    table = 'listens_{}'.format(datetime.strftime(date, '%Y_%m'))
+    print(table)
+    df.registerTempTable(table)
```
mayhem: Could this fail? If so, it should be in a try block.

listenbrainz_spark/stats/user.py
mayhem: Yes, anytime that something complex such as this is carried it out, it should have a try block around it. If you cannot envisions a problem during such a complex operation, use a generic try block that then prints a stack track and attempts to carry on if it can.

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,214 @@
...
+        data[user_name]['artists'] = {
+            'artist_stats': artist_stats,
+            'artist_count': len(artist_stats),
+        }
+
+    recording_data = get_recordings(table)
+    for user_name, recording_stats in recording_data.items():
+        data[user_name]['recordings'] = recording_stats
+
+    release_data = get_releases(table)
```
mayhem: Here and through the rest of the file, the same try block should be used.

listenbrainz_spark/stats/user.py
```diff
@@ -0,0 +1,214 @@
...
+    for user_name, metadata in data.items():
+        data[user_name]['yearmonth'] = yearmonth
+        rabbitmq_data = {
+            'type' : 'user',
+            user_name : metadata,
+        }
+        try:
+            rabbbitmq_conn_obj.start(rabbitmq_data)
+            print("Statistics of %s pushed to rabbitmq" % (user_name))
+        except Exception as err:
```
mayhem: We have a lot better understanding what type of error might be thrown here. Please read the docs and see what exceptions might be thrown during start()

listenbrainz_spark/stats_writer/stats_writer.py
```diff
@@ -0,0 +1,75 @@
...
+        while True:
+            try:
+                credentials = pika.PlainCredentials(username, password)
+                connection_parameters = pika.ConnectionParameters(
+                    host=host,
+                    port=port,
+                    virtual_host=virtual_host,
+                    credentials=credentials,
+                )
+                self.connection =  pika.BlockingConnection(connection_parameters)
```
mayhem: Here too, we can find more specific errors to catch. See the LB main code for examples.

listenbrainz_spark/stats_writer/stats_writer.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        try:
+            self.unique_ch.basic_publish(
+                exchange=config.SPARK_EXCHANGE,
+                routing_key='',
+                body=json.dumps(data),
+                properties=pika.BasicProperties(delivery_mode = 2,),
+            )
+            return True
+        except pika.exceptions.ConnectionClosed:
```
mayhem: This is a good except block!

spark-submit.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+zip -r listenbrainz_spark.zip listenbrainz_spark/
+time ./run.sh /usr/local/spark/bin/spark-submit --master spark://spark-master.spark-network:7077 --num-executors=28 --executor-memory=1g --driver-memory=1g --py-files listenbrainz_spark.zip "$@"
```
mayhem: Before too long we will want to move these into a config file. For instance, once we move to azure, these values may no longer fit.

listenbrainz_spark/stats/user.py
vansika: Not really, but a general try block would be good, I guess.

listenbrainz_spark/stats/user.py
vansika: After catching exception it is preferable to stop the script because there is a specific format in which data should be sent to LB server, any irregularity in that should terminate the script.

spark-submit.sh
vansika: The values will be moved to config once they are finalised.

listenbrainz_spark/stats_writer/stats_writer.py
vansika: Thanks :)

spark-submit.sh
mayhem: Can you please make a jira ticket for this, so it doesn't slip off our radar?

paramsingh:
>I'm still unhappy about the blanket exceptions. How do you feel @paramsingh?

I think we should merge this for now. I'll take the exceptions work over.

PR #23, opened 2019-03-03
creating dataframe from .parquet stored in hdfs

vansika:
Creating data frames from files stored in HDFS. We shall not use data dumps directly.


create_dataframes.py
```diff
@@ -0,0 +1,40 @@
...
+             From listen
+        GROUP BY user_name, track_name, recording_msid
+        ORDER BY cnt desc
+        """)
+    return users_df
+
+if __name__ == '__main__':
+    listenbrainz_spark.init_spark_session('Create_Datafeame')
+    df = None
+    for y in range(2005, 2020):
```
mayhem: The first year could be hardcoded, but the end year should not be.

create_dataframes.py
```diff
@@ -0,0 +1,40 @@
...
+        for m in range(1, 13):
+            try:
+                month = listenbrainz_spark.sql_context.read.parquet('{}/data/listenbrainz/{}/{}.parquet'.format(config.HDFS_CLUSTER_URI, y, m))
+                df = df.union(month) if df else month
+            except:
+                print("No listens for {}/{}".format(m,y))
+                continue
+    df.printSchema()
+    print("Registering Dataframe...")
+    df.registerTempTable('listen')
```
mayhem: I think we need to use better table names so we don't get ourselves confused...

vansika:
- The input to ALS is an RDD[Rating] and Rating is an (int, int, double). To add a new column to existing data frames (user_id, recording_id) I came across two methods. The first one is to add a new column using 'withColumn', 'monotonically_increasing_id()" would be used to get distinct 64-bit integers which won't be consecutive. This would throw error since rating does not accept long int and hence we may have to use hashing to get around with it (https://stackoverflow.com/questions/30314879/how-to-use-long-user-id-in-pyspark-als). The other approach is to use 'zipWithIndex()' which will convert the data frame to a rdd , add the column and then we would have to convert the rdd to a data frame. I am not sure which of the two will have more overhead but I believe the former is better because we won't have listens confined to simply 'int' (2,147,483,647).
- Presently,  load_data/create_dataframe is fetching data from hdfs and then storing intermediate tables in directory, train_model is using this directory to create one more directory and finally recommend.py is using the directory to print recommendations. Should we store every intermediate data directory and recommendations in HDFS. Also, do we need to override existing tables with each cycle of recommendation or create new ones? I am a little confused about how often the file would be processed. 
- Can we consider the table name? Maybe we can give a better name. right now it is of the format: (df_to_train_year_month_date)
- Like load_data was throwing empty rdd error for two users, create_dataframes did not but the recommendations were not good, for sure. I am not sure why it did not throw any error. Well, the best would be to test on the whole data. The next step for me would be that, for sure.

mayhem:
Regarding withColumn/zipWithIndex... I think we need to create a final column that uses double as a rating -- I don't think it can convert the type for us. We need to check this. But a 2^32-1 rating is fine -- certainly for now. I would suggest that you test both approaches and see which one runs faster.

mayhem:
Regarding data lifecycles:

- The intermediate tables that might get created in order to create the final table that we can feed to CF should get removed before this script ends. Where the intermediate files are created/live does not matter.
- The final table should be stored in HDFS in a location on HDFS that makes it clear it belongs to the recommendation stuff. Tables should be named with dates in them -- what you have might be ok for now. 
- The load_data/process data script will be run when we want to update the input tables for the CF algorithm. Let's say that we decide to update it weekly -- then we will want run that script weekly and if we re-train the data, then we will simply re-use the tables that are current for this week.
- In order to generate recommendations we will need to specify the location of a model. This model will be used to then generate a recommendation. The results of this recommendation are not going to be stored in in HFDS, but likely written to another RabbitMQ queue for shipping off to the LB server.

vansika:
print in train_models.py, line 27 is redundant.

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
...
+import os, sys
+import train_models
+import recommend
+
+def prepare_user_data(table):
+    users_q = spark.sql("""
+            SELECT DISTINCT user_name
+             From %s
+        """ % (table))
+    users_rdd = users_q.rdd.zipWithIndex()
```
mayhem: If there is any way we can do this step as part of the SQL query, that would be ideal.

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
...
+    recordings_df = spark.createDataFrame(recordings_rdd)
+    return recordings_df
+
+def get_playcounts_data(listens_df, users_df, recordings_df):
+
+    listens_df.createOrReplaceTempView('listen')
+    users_df.createOrReplaceTempView('user')
+    recordings_df.createOrReplaceTempView('recording')
+    playcounts_df = spark.sql("""
+        SELECT user_id,
```
mayhem: This is the place where we can do integer conversions -- this way we can skip several steps. Discussed in IRC, so I'll stop for now.

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
...
+      GROUP BY user_id, recording_id
+      ORDER BY user_id
+    """)
+    return playcounts_df
+
+
+
+if __name__ == '__main__':
+    listenbrainz_spark.init_spark_session('Create_Dataframe')
+    spark = SparkSession.builder.getOrCreate()
```
paramsingh: Why does this need to be created? Why not use listenbrainz_spark.session?

listenbrainz_spark/__init__.py
```diff
@@ -1,19 +1,13 @@
...
-    global session, context, sql_context
+    global session
     session = SparkSession \
             .builder \
             .appName(app_name) \
             .config("spark.hadoop.dfs.client.use.datanode.hostname", "true") \
             .config("spark.hadoop.dfs.datanode.use.datanode.hostname", "true") \
             .getOrCreate()
-    context = session.sparkContext
-    context.setLogLevel("ERROR")
```
paramsingh: Why are we removing this?

I'm pretty sure context and sql_cobtext are used elsewhere

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
+import listenbrainz_spark
+from listenbrainz_spark import config
+from pyspark.sql import Row, SparkSession
+from datetime import datetime
+import os, sys
+import train_models
+import recommend
+
+def prepare_user_data(table):
+    users_q = spark.sql("""
```
paramsingh: I dont see spark being passed anywhere?

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
...
+                 , user_name
+             From %s
+        """ % (table))
+    return listens_df
+
+def prepare_recording_data(table):
+    recordings_q = spark.sql("""
+            SELECT track_name
+                 , recording_msid
+             From %s
```
paramsingh: Capital FROM?

create_dataframes.py
```diff
@@ -0,0 +1,96 @@
...
+    return playcounts_df
+
+
+
+if __name__ == '__main__':
+    listenbrainz_spark.init_spark_session('Create_Dataframe')
+    spark = SparkSession.builder.getOrCreate()
+    date = datetime.utcnow()
+    df = None
+    for y in range(2005, date.year+1):
```
paramsingh: 2002? Also this really should be a constant in some file

recommend.py
```diff
@@ -1,8 +1,16 @@
...
+import tempfile
+
+from pyspark.mllib.recommendation import MatrixFactorizationModel
+from pyspark import SparkContext
+from listenbrainz_spark import hdfs_connection, config
+from datetime import datetime
+from pyspark.sql import SparkSession
 
+spark = None
+sc = None
```
paramsingh: These should come from the listenbrainz_spark module ideally

vansika:
@mayhem changes in the script according to the proposed guide will reflect in a new pull request.

PR #24, opened 2019-03-21
Docker

vansika:
Dependencies stated in requirements.txt were not getting installed on master and workers since the path to 
requirments.txt was not mentioned in the docker files.

### DIFF
--- a/Dockerfile.master
+++ b/Dockerfile.master
@@ -14,6 +14,8 @@ RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
 
 RUN mkdir /rec
 WORKDIR /rec
+COPY requirements.txt /rec/requirements.txt
+RUN pip3 install -r requirements.txt
 COPY . /rec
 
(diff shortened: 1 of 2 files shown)


vansika:
@mayhem In case we need to install any package on master as well as worker nodes, we can update that in requirements.txt and simply build.  I am not sure what do you mean by python-files here.

mayhem:
By python-files I mean python modules. I don't see how these containers need python support -- we can add it when we need it -- otherwise this just bloats our images...

vansika:
Any module should be installed on master as well as workers. Instead of making changes in Dockerfile.master and worker separately , we can just update requirements.txt. A few days ago, I hit 'no module hdfs' error since it was not installed on workers, even if it was mentioned in requirements.txt because the dockerfiles did not have command to execute requirements.txt. 

What you say makes sense, it's just adding modules at two different files troubles me.

mayhem:
Well, this can't hurt, I'll just merge it.

PR #25, opened 2019-04-07
Artist popularity from MLHD.

paramsingh:
yo yo


listenbrainz_spark/mlhd/scripts/artist_popularity.py
```diff
@@ -0,0 +1,30 @@
+import os
+import listenbrainz_spark
+from datetime import datetime
+
+from listenbrainz_spark import config
+from listenbrainz_spark.mlhd import MLHD_DATA_PATH
+
+
+def main():
```
mayhem: Overall there are exception handling missing and we may want to choose a different data format for delivering the data (i.e. for inserting into and keeping a table inside a new schema inside the MB database). But we're not certain that this data will be useful, so we let's not delve into that right now.

spark-submit.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
 zip -r listenbrainz_spark.zip listenbrainz_spark/
-time ./run.sh /usr/local/spark/bin/spark-submit --master spark://spark-master.spark-network:7077 --num-executors=28 --executor-memory=1g --driver-memory=1g --py-files listenbrainz_spark.zip "$@"
-
+time ./run.sh /usr/local/spark/bin/spark-submit \
+	--packages org.apache.spark:spark-avro_2.11:2.4.1 \
+	--master spark://spark-master.spark-network:7077 \
+	--num-executors=7 \
```
mayhem: I dislike how these cluster specific things are hardcoded. That is not only a problem here but in other scripts as well. We need to start having a cluster dependent config file for these config values...

paramsingh:
Ready for review again.

paramsingh:
>I wonder how many files we can run on our existing cluster. Really time to play with azure...

I ran it on 50 files, which is around 2 billion listens, took around 8 min.

PR #31, opened 2019-06-21
candidate sets for each user to personalize recommendations

vansika:
As per IIRC conversation, for each user, two candidate sets are generated consisting of recording ids of top artists and similar artists.


listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+
+    listens_df = None
+    for y in range(config.STARTING_YEAR, config.ENDING_YEAR + 1):
+        for m in range(config.STARTING_MONTH, config.ENDING_MONTH + 1):
+            try:
+                month = listenbrainz_spark.sql_context.read.parquet('{}/data/listenbrainz/{}/{}.parquet'
+                    .format(config.HDFS_CLUSTER_URI, y, m))
+                listens_df = listens_df.union(month) if listens_df else month
+            except AnalysisException as err:
+                logging.error('Cannot read parquet files from HDFS: {} \n {}'.format(type(err).__name__,str(err)))
```
mayhem: I think the error message should state which files could not be read.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+        sys.exit(-1)
+    if not artists_relation_df:
+        raise SystemExit("Parquet file conatining artist-artist relation is missing from HDFS. Aborting...")
+
+    try:
+        path = os.path.join(config.HDFS_CLUSTER_URI, 'data', 'listenbrainz', 'recommendation-engine', 'dataframes')
+        recordings_df = listenbrainz_spark.sql_context.read.parquet(path + '/recordings_df.parquet')
+        users_df = listenbrainz_spark.sql_context.read.parquet(path + '/users_df.parquet')
+    except AnalysisException as err:
+        logging.error('Cannot read parquet files from HDFS: {} \n {}'.format(type(err).__name__,str(err)))
```
mayhem: Same here, please be more specific in the error.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+                else:
+                    df = get_similar_artists_without_limit(top_artists)
+                if not df.take(1):
+                    logging.info('Candidate sets cannot be generated since no similar artists for any of the \
+                        top artists found: \n{}'.format(top_artists))
+                    continue
+                similar_artists_df = get_similar_artists_with_limit(df)
+
+                # Gets only those top artists whose similar artists count is not 0
+                top_artist_with_collab_df = df.select('artist_name_0').distinct()
```
mayhem: I guess even here we need to take into account artist_name_1, right?

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+                }
+    """
+    date = datetime.utcnow().strftime('%Y-%m-%d')
+    candidate_html = 'Canidate-{}-{}.html'.format(uuid.uuid4(), date)
+    context = {
+        'user_data' : user_data
+    }
+    utils.save_html(candidate_html, context, 'candidate.html')
+
+def main():
```
mayhem: This main function is really quite long. If it can be broken up into smaller functions, that would be good -- it is hard to read as it is.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+    user_data = defaultdict(dict)
+    with open(path) as f:
+        recommendation_metadata = json.load(f)
+        similar_artists_candidate_set_df = None
+        top_artists_candidate_set_df = None
+        for user_name in recommendation_metadata['user_name']:
+            top_artists = ()
+            similar_artists = ()
+            top_artists_with_collab = ()
+            try:
```
paramsingh: This try block is probably a bit too large, debugging problems in here would not be easy, we should break it into smaller try catch blocks with different log messages.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+    except Exception as err:
+        logging.error('An error occurred: {} \n {}. Aborting...'.format(type(err).__name__,str(err)), exc_info=True)
+        sys.exit(-1)
+
+    listens_df = None
+    for y in range(config.STARTING_YEAR, config.ENDING_YEAR + 1):
+        for m in range(config.STARTING_MONTH, config.ENDING_MONTH + 1):
+            try:
+                month = listenbrainz_spark.sql_context.read.parquet('{}/data/listenbrainz/{}/{}.parquet'
+                    .format(config.HDFS_CLUSTER_URI, y, m))
```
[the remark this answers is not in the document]
vansika: I have pushed the code after running. I will include a backslash anyway :)

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+                            'top_artists 1' : ['similar_artist 1', 'simialr_artist 2' ... 'similar_artist x'],
+                        ...
+                        'top_artists y' : ['similar_artist 1', 'simialr_artist 2' ... 'similar_artist x'],
+                        },
+                        'time' : 'xxx'
+                    },
+                }
+    """
+    date = datetime.utcnow().strftime('%Y-%m-%d')
+    candidate_html = 'Canidate-{}-{}.html'.format(uuid.uuid4(), date)
```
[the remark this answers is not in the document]
vansika: thanks :)

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+            artists (tuple): A tuple of artist names.
+            user_id (int): User id of the user.
+
+        Returns:
+            candidate_recording_ids_df (dataframe): Dataframe with columns as:
+                ['user_id', 'recording_id']
+    """
+    df = run_query("""
+        SELECT recording_id
+            FROM recording
```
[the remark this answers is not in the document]
vansika: How is it off?  There is a tab before 'FROM'. The link was not of much help, maybe I am not able to see a difference.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+         WHERE artist_name IN %s
+    """ % (artists,))
+    candidate_recording_ids_df = df.withColumn('user_id', lit(user_id)).select('user_id', 'recording_id')
+    return candidate_recording_ids_df
+
+def get_user_id(user_name):
```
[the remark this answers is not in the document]
vansika: This should be shifted to utils, I guess. For recommend.py I can open a separate PR.

listenbrainz_spark/recommendations/candidate_sets.py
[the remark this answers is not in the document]
paramsingh: It isn't about the tabs, the SELECT and FROM need to be right aligned.

```
SELECT recording_id
  FROM recording
 WHERE artist_id IN %s
```

listenbrainz_spark/recommendations/candidate_sets.py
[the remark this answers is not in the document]
paramsingh: I don't think merging PRs with copied over functions is a good idea, just import it from recommend.py

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,363 @@
...
+
+def save_candidate_html(user_data):
+    """ Save user data to an HTML file.
+
+        Args:
+            user_data (dict): Dcitionary can be depicted as:
+                {
+                    'user_name 1': {
+                        'artists': {
+                            'top_artists 1' : ['similar_artist 1', 'simialr_artist 2' ... 'similar_artist x'],
```
paramsingh: spelling of `similar` here and in other docstrings is incorrect too.

listenbrainz_spark/recommendations/candidate_sets.py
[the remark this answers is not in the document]
vansika: There are many functions that are copied over and over again in create_dataframes.py, recommend.py and train_models.py. 
They can be included in utils and imported from there like the one above.

listenbrainz_spark/recommendations/candidate_sets.py
vansika: Thank you for pointing out.

listenbrainz_spark/recommendations/candidate_sets.py
vansika: I have checked the error. 'str(err)' displays the path of files which could not be read, therefore I did not include them.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,384 @@
...
+        Raises:
+            Py4JJavaError if candidate recording ids query cannot be parsed.
+    """
+    similar_artists = ()
+    # eliminate artists from similar artists who are a part of top artists
+    similar_artists_df = similar_artists_df.select('similar_artist_name').subtract(top_artists_df.select('artist_name'))
+    if not similar_artists_df.take(1):
+        logging.info('Similar artists candidate set not generated for "{}" as similar artists \
+            are equivalent to top artists for the user'.format(user_name))
+        return False
```
paramsingh: Instead of returning True False in a function which also returns dataframes, we should raise exceptions.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,384 @@
...
+    try:
+        if len(top_artists_with_collab) == 1:
+            top_artists_recording_ids_df = get_candidate_recording_ids(tuple((
+                top_artists_with_collab[0])),user_id)
+        else:
+            top_artists_recording_ids_df = get_candidate_recording_ids(top_artists_with_collab, user_id)
+    except Py4JJavaError as err:
+        logging.error('An error occured while fetching recording ids of top artists for "{}". Candidate set \
+            cannot be generated: {} \n {}.'.format(user_name, type(err).__name__, str(err)))
+        return False
```
paramsingh: Same here

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,384 @@
...
+        Returns:
+            similar_artists_recording_ids_df (dataframe): Dataframe with columns as:
+                ['user_id', 'recording_id']
+
+        Raises:
+            Py4JJavaError if candidate recording ids query cannot be parsed.
+    """
+    similar_artists = ()
+    # eliminate artists from similar artists who are a part of top artists
+    similar_artists_df = similar_artists_df.select('similar_artist_name').subtract(top_artists_df.select('artist_name'))
```
paramsingh: Instead of using the DataFrame API, we should use SQL here (we prefer SQL over the ORM in all our other projects too)

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,384 @@
...
+                    'top_artists 2' : ['similar_artist 1', 'similar_artist 2' ... 'similar_artist x'],
+                    .
+                    .
+                    .
+                    'top_artists y' : ['similar_artist 1', 'similar_artist 2' ... 'similar_artist x'],
+                }
+    """
+    artists = defaultdict(dict)
+    top_artist_with_collab_df = similar_artist_df.select('artist_name').distinct()
+    for row in top_artist_with_collab_df.collect():
```
paramsingh: Should be done in SQL.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -0,0 +1,384 @@
...
+    candidate_html = 'Candidate-{}-{}.html'.format(uuid.uuid4(), date)
+    context = {
+        'user_data' : user_data
+    }
+    utils.save_html(candidate_html, context, 'candidate.html')
+
+def main():
+    ti = time()
+    if not utils.initialize_spark_session('Candidate_set'):
+        raise SystemExit('Aborting...')
```
paramsingh: Generally, try catch is considered more pythonic over this `if not` pattern

listenbrainz_spark/recommendations/utils.py
```diff
@@ -9,3 +15,40 @@ def save_html(filename, context, template):
     with open(outputfile, 'w') as f:
         html = template_environment.get_template(template).render(context)
         f.write(html)
+
+def initialize_spark_session(app_name):
```
paramsingh: Instead of creating a new function, just make these changes in the `init_spark_session` function in  listenbrainz_spark/__init__.py

paramsingh: Oh and raise the exception instead of returning bool

listenbrainz_spark/recommendations/utils.py
```diff
@@ -9,3 +15,40 @@ def save_html(filename, context, template):
...
+        logging.error('Cannot initialize Spark session "{}": {} \n {}.'.format(app_name, type(err).__name__,
+            str(err)), exc_info=True)
+        return False
+    except Exception as err:
+        logging.error('An error occurred while initializing Spark session "{}": {} \n {}.'.format(app_name,
+            type(err)  .__name__,str(err)), exc_info=True)
+        return False
+
+def register_dataframe(df, table_name):
+    try:
```
paramsingh: This function is useful for stats also, let's put it inside listenbrainz_spark/__init__.py or listenbrainz_spark/utils.py

listenbrainz_spark/recommendations/utils.py
```diff
@@ -9,3 +15,40 @@ def save_html(filename, context, template):
...
+        return True
+    except AnalysisException as err:
+        logging.error('Cannot register dataframe "{}": {} \n {}.'.format(table_name, type(err).__name__, str(err)))
+        return False
+    except Exception as err:
+        logging.error('An error occured while registering dataframe "{}": {} \n {}.'.format(table_name,
+            type(err).__name__, str(err)), exc_info=True)
+        return False
+
+def read_files_from_HDFS(path):
```
paramsingh: Should also be moved to listenbrainz_spark/init.py

vansika:
@mayhem Pyspark errors generally fall under py4j module (https://www.py4j.org/py4j_java_protocol.html) which are verbose and should be handled carefully.
Pyspark queries throw py4j.protocol.Py4JJavaError exceptions which are the main class of exceptions to (AnalysisException etc...). To track the origin of exception, these subclasses of Py4JJavaError exceptions should be traced. 
Other Spark error such as network errors etc can also be traced using py4j. I have tried to catch all the possible exceptions which I could reach out to by writing bad smelling code,  yet the PR does not cover all the possible exceptions.
When it would be run over and over again on different datasets and different users in the testing period, we can hopefully trace all the exceptions.

vansika:
@paramsingh  I have tried to cover all the areas of improvement pointed out by you. I will be grateful to know if there is anything else I should work on.

PR #33, opened 2019-06-29
upgrade spark version from 2.4.1 to 2.4.3

vansika:
Perhaps, Spark tar files are frequently uploaded and the old ones are deleted here: http://apachemirror.wuchna.com/spark/ 

http://apachemirror.wuchna.com/spark/spark-2.4.1/spark-2.4.1-bin-hadoop2.7.tgz returns 404 not found.


vansika:
Due to some issues on my local machine, it was not able to resolve https://archive.apache.org/dist/.
Everything works fine now with the older version :)

PR #35, opened 2019-07-06
Add scripts to download and apply incremental dumps 

paramsingh:
Now that we have incremental dumps, I've written a script `import_incremental.py` that we can use to download and apply incremental dumps, starting from the current version of the data.


import_full.py
```diff
@@ -0,0 +1,25 @@
...
+from listenbrainz_spark import hdfs_connection
+from listenbrainz_spark.data.import_data import import_dump
+from listenbrainz_spark.data.import_data.ftp import ListenBrainzFTPDownloader
+
+if __name__ == '__main__':
+    if len(sys.argv) != 2:
+        print("Usage: import.py <app_name>")
+        sys.exit(-1)
+
+    listenbrainz_spark.init_spark_session(sys.argv[1])
```
mayhem: We need to catch exceptions and print nice error messages at the top level. And make sure to use good exit codes and practices set forth by https://github.com/metabrainz/metabrainz-howto-guides/blob/master/writing_effective_data_processing_scripts.md

paramsingh:
This won't be needed once spark-writer comes back.

PR #36, opened 2019-07-11
Use a single writer script for recommendations and stats writer

vansika:
A single *writer* script can be used to write stats and recommendations to different queues and exchanges. 
We can think of renaming stats_writer.py for it can write any type of data(stats, recommendations, etc) to the given queue.


vansika:
@mayhem yes, absolutely. I mentioned the file name but missed the class in description. Oops.

The name sounds good to me :)

PR #38, opened 2019-07-17
Train model on last X months listens

vansika:
This PR aims at loading listens of past X months to train the data and use the model for recommendations, where X is a config value.

### DIFF
--- a/listenbrainz_spark/recommendations/create_dataframes.py
+++ b/listenbrainz_spark/recommendations/create_dataframes.py
@@ -8,6 +8,7 @@ from py4j.protocol import Py4JJavaError
 
 import listenbrainz_spark
 from listenbrainz_spark import config, utils
+from listenbrainz_spark.stats import adjusted_date
 from listenbrainz_spark.exceptions import SQLException
 from listenbrainz_spark.recommendations.utils import save_html
 from listenbrainz_spark.sql import create_dataframes_queries as sql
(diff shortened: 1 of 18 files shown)


listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -26,19 +27,41 @@ def save_dataframe_html(users_df_time, recordings_df_time, playcounts_df_time, t
...
     context = {
         'users_df_time' : users_df_time,
         'recordings_df_time' : recordings_df_time,
         'playcounts_df_time' : playcounts_df_time,
-        'lb_dump_time_window' : lb_dump_time_window,
         'total_time' : total_time
     }
     save_html(queries_html, context, 'queries.html')
 
+def training_data_window():
```
mayhem: I find this function very confusing. It seems that some functionality (merging dfs) is duplicated here and in utils.get_listens(). Also, the date math in this function seems complicated. I think it might be easier to extend the functionality of get_listens to work across years and this function should really only do date calculations.

listenbrainz_spark/utils.py
```diff
@@ -42,15 +43,20 @@ def read_files_from_HDFS(path):
...
     except AnalysisException as err:
-        raise AnalysisException('Cannot read "{}" from HDFS: {}\n'.format(path, type(err).__name__),
+      raise AnalysisException('Cannot read "{}" from HDFS: {}\n'.format(path, type(err).__name__),
             stackTrace=traceback.format_exc())
     except Py4JJavaError as err:
         raise Py4JJavaError('An error occurred while fetching "{}": {}\n'.format(path, type(err).__name__),
             err.java_exception)
 
-def get_listens():
+def get_listens(d1, d2):
```
mayhem: Function declarations should really have better variable names. Also the args and the docstring do not agree (m1 vs d1)

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -26,19 +27,34 @@ def save_dataframe_html(users_df_time, recordings_df_time, playcounts_df_time, t
...
```
mayhem: Not sure I really like the name of this function. create_training_data_from_window() ?

listenbrainz_spark/utils.py
mayhem: Also, from this function it is unclear if the m2 month will be included. The < says it won't, but this isn't clear from the documentation.

listenbrainz_spark/utils.py
```diff
@@ -49,13 +49,12 @@ def read_files_from_HDFS(path):
         raise Py4JJavaError('An error occurred while fetching "{}": {}\n'.format(path, type(err).__name__),
             err.java_exception)
 
-def get_listens(d1, d2):
+def get_listens(current_date, adjusted_date):
```
mayhem: In this context current_date and adjusted_date make very little sense -- that context comes from the calling functions. begin_date, end_date? The docs should say if the end date is included in the results or not.

listenbrainz_spark/utils.py
```diff
@@ -64,13 +63,15 @@ def get_listens(d1, d2):
                     'recording_msid', 'release_mbid', 'release_msid', 'release_name', 'tags',
                     'track_name', 'user_name'
                 ]
+        Note: Dataframe of current month would not be included.
```
vansika: @mayhem The note here indicates that the current date/end date would not be included.

listenbrainz_spark/utils.py
vansika: okay. begin_date and end_date seem fine to me. Should I change the names in create_training_data_from_window also?

mayhem: maybe get_listens_for_date_range() ?

vansika: Yes. Cool.

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -49,10 +49,9 @@ def training_data_window():
         Note: Under the assumption that config.TRAIN_MODEL_WINDOW will always indicate months.
     """
     training_df = None
-    # go to first of the current month and then decrement months
-    d1 = stats.replace_days(1)
-    d2 = stats.adjust_months(d1, -config.TRAIN_MODEL_WINDOW)
-    training_df = utils.get_listens(d2, d1)
+    current_date = datetime.utcnow()
+    adjusted_date = stats.adjust_months(current_date, -config.TRAIN_MODEL_WINDOW)
+    training_df = utils.get_listens(current_date, adjusted_date)
```
vansika: should I change current_date and adjusted_date to begin_date and end_date here also?

PR #40, opened 2019-07-30
Weekly dataframe to get top-artists for generating recommendations

vansika:
We intend to generate recommendations after every X days. This PR fetches dataframes to look for listens which are listened to during these X days.


listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -35,7 +35,12 @@ def get_listens_for_rec_generation_window():
     df = None
     end_date = datetime.utcnow()
     begin_date = stats.adjust_days(end_date, -config.RECOMMENDATION_GENERATION_WINDOW)
```
vansika: @mayhem  If everything looks okay, should I introduce a check for config.REOMMENDATION_GENERATION_WINDOW (if not less than 28 abort)

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -20,6 +21,28 @@
...
+                    artist_mbids, artist_msid, artist_name, listened_at, recording_mbid,
+                    recording_msid, release_mbid, release_msid, release_name, tags, track_name, user_name
+                ]
+        Note: Under the assumption that config.RECOMMENDATION_GENERATION_WINDOW will always indicate days.
+    """
+    df = None
+    end_date = datetime.utcnow()
+    begin_date = stats.adjust_days(end_date, -config.RECOMMENDATION_GENERATION_WINDOW)
+    # config.RECOMMENDATION_GENERATION_WINDOW should be less than 28 (considering leap year)
+    # otherwise the function may produce unexpected results (skip a few months data).
```
mayhem: If this is the problems, then we need to do better date/time math.

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -37,7 +37,7 @@ def save_dataframe_html(users_df_time, recordings_df_time, playcounts_df_time, t
     }
     save_html(queries_html, context, 'queries.html')
 
-def create_training_data_from_window():
+def get_listens_for_train_model_window():
```
mayhem: get_listens_for_training_model_window would be more idiomatic english.

mayhem: And you two the same function name here and in the utils module -- they differ and they should have different names, no?

listenbrainz_spark/stats/__init__.py
```diff
@@ -44,3 +44,7 @@ def replace_days(day):
 def adjust_months(date, months):
```
mayhem: Both of these functions bother me a little in that they are not very flexible. What if we didn't want a date range that led up to now? What if we wanted listens between arbitrary dates?

listenbrainz_spark/utils.py
```diff
@@ -49,31 +49,76 @@ def read_files_from_HDFS(path):
...
+            df (dataframe): Columns can be depicted as:
+                [
+                    'artist_mbids', 'artist_msid', 'artist_name', 'listened_at', 'recording_mbid'
+                    'recording_msid', 'release_mbid', 'release_msid', 'release_name', 'tags',
+                    'track_name', 'user_name'
+                ]
+        Note: Listens of current month will be fetched.
+    """
+    df = None
+    while begin_date <= end_date:
```
mayhem: This is also very unclear to me what is happening here.

listenbrainz_spark/stats/__init__.py
vansika: IMO, listens between arbitrary date range can be fetched by changing the begin_date and end_date, these functions just serve the purpose of going back and ahead in time.

listenbrainz_spark/utils.py
```diff
@@ -49,29 +49,32 @@ def read_files_from_HDFS(path):
         raise Py4JJavaError('An error occurred while fetching "{}": {}\n'.format(path, type(err).__name__),
             err.java_exception)
 
-def get_listens_for_date_range(begin_date, end_date):
-    """ Loads all the listens listened to in a given time window from HDFS.
+def get_listens(from_date, to_date):
```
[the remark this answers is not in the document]
vansika: Nice thinking :)

listenbrainz_spark/stats/__init__.py
```diff
@@ -37,10 +37,20 @@ def run_query(query):
...
 
-def replace_days(day):
-    date = datetime.utcnow().replace(day=day)
+def replace_days(date, day):
+    date = date.replace(day=day)
     return date
 
-def adjust_months(date, months):
-    date = date + relativedelta(months=months)
+def adjust_months(date, months, shift_backwards=True):
```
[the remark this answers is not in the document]
vansika: I am keeping it as it is for now as per the IRC conversation.

PR #42, opened 2019-08-01
Schema to store model metadata in HDFS

vansika:
This PR aims at creating a schema for dataframe which will store metadata of every best model trained so that we have important information about best models even after they are deleted from HDFS.


vansika:
@mayhem should we add a column -> 'total_listens' which will be a summation of listens used to train, validate and test the model. That way we don't have to manually add them to get the total number of listens. What do you think? 
Is there any other field(column) that you would like to add?

mayhem:
I think total_listens is a good idea. num_iterations and alpha would also be useful, even those those are constant now. We'll want to change those as we get more picky.

listenbrainz_spark/schema.py
```diff
@@ -19,12 +19,30 @@
     StructField('user_name', StringType(), nullable=False),
 ]
 
+model_metadata_schema = [
+    StructField('alpha', FloatType(), nullable=False), # Baseline level of confidence weighting applied.
+    StructField('begin_date', TimestampType(), nullable=False), # Timestamp from when listens have been used to train, validate and test the model.
+    StructField('created', TimestampType(), nullable=False), # Timestamp when the model is saved in HDFS.
+    StructField('deleted', TimestampType(), nullable=True), # Timestamp when the model is deleted from HDFS.
+    StructField('end_date', TimestampType(), nullable=False), # Timestamp till when the listens have been used to train, validate and test the model.
+    StructField('lmbda', FloatType(), nullable=False), # Controls over fitting.
```
mayhem: Was this spelling intentional?

PR #43, opened 2019-08-11
HDFS utilities for modules and scripts

vansika:
This PR intends to merge some HDFS utilities in the codebase to facilitate future PRs

### DIFF
--- a/listenbrainz_spark/utils.py
+++ b/listenbrainz_spark/utils.py
@@ -105,7 +105,7 @@ def create_dir(path):
             path (string): Path of the directory to be created.
 
         Note: >> Caller is responsibe for initializing HDFS connection.
-              >> The function do not throws an error if the directory already exists.
+              >> The function does not throw an error if the directory path already exists.
     """
     hdfs_connection.client.makedirs(path)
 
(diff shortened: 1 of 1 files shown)


listenbrainz_spark/utils.py
```diff
@@ -96,3 +97,35 @@ def save_parquet(df, path):
...
+        Args:
+            path (string): Path of the directory to be created.
+
+        Note: >> Caller is responsibe for initializing HDFS connection.
+              >> The function do not throws an error if the directory already exists.
+    """
+    hdfs_connection.client.makedirs(path)
+
+def delete_dir(path):
+    """ Deletes a directory recursively from HDFS.
```
mayhem: Recursive delete can be dangerous. Can you please makes the default false?

listenbrainz_spark/utils.py
```diff
@@ -96,3 +97,35 @@ def save_parquet(df, path):
...
+
+        Args:
+            path (string): Path of the directory to be deleted.
+
+        Note: Caller is responsible for initializing HDFS connection.
+    """
+    hdfs_connection.client.delete(path, recursive=True)
+
+def get_status(path):
+    """ Checks the status of a directory in HDFS. May be used to check if a directory
```
mayhem: What is the return value of this?

vansika: https://hdfscli.readthedocs.io/en/latest/api.html#hdfs.client.Client.makedirs

vansika: If you try to get the return value in a variable and print, the value would always be 'None'

vansika: http://hadoop.apache.org/docs/r1.0.4/webhdfs.html#FileStatus

vansika: If the directory does not exist, it will throw HDFS error. If we catch an error, it means no directory with such name exists, and hence we can call create_dir because  this -https://hdfscli.readthedocs.io/en/latest/api.html#hdfs.client.Client.makedirs  
does not raise an error if we try to create a directory with a path that already exists.

mayhem: You should put that into the docstrings, not to me. :)

PR #44, opened 2019-08-11
Modify develop.sh to use a single command for building services and starting containers

vansika:
Presently, we need to issue two commands: 
- develop.sh down to stop the running containers
- develop.sh up to start the containers

With this PR only one command ( develop.sh ) would be enough to build and start all the services.


mayhem:
agreed, I don't quite see the point of this...

vansika:
Good to close :)

PR #45, opened 2019-08-15
A single file to prepare HDFS to contain recommendation data and other utilities.

vansika:
I have skipped '/data/listenbrainz' and '/data/listenbrainz/similar_artist' since both the directories must be pre-filled with data for recommendation engine to work. In time when we have scripts for uploading data in these directories, we can include them in manage.py and call the data uploading scripts.


manage.py
```diff
@@ -1,29 +1,107 @@
...
+from hdfs.util import HdfsError
+from py4j.protocol import Py4JJavaError
+
+@click.group()
+def cli():
+    pass
+
+@cli.command(name='init_dir')
+@click.option('--rm', is_flag=True, help='Delete existing directories from HDFS.')
+@click.option('--f', is_flag=True, help='Delete existing directories from HDFS recursively.')
```
mayhem: -f suggests "force". I think --recursive might be a better option here. Harder to get wrong too, which is more safe for us.

manage.py
```diff
@@ -1,29 +1,107 @@
...
+            utils.delete_dir(path.CHECKPOINT_DIR)
+            logging.info('Successfully deleted directories.')
+        except HdfsError as err:
+            logging.error('{}: Some/all directories are non-empty. Try "--f" to delete recursively.'.format(
+                type(err).__name__))
+            logging.warning('Deleting directory recursively will delete all the recommendation data.')
+            sys.exit(-1)
+
+    if f:
+        utils.delete_dir(path.RECOMMENDATION_PARENT_DIR, recursive=True)
```
mayhem: Why no exception checking here? Things can still go wrong!

manage.py
```diff
@@ -1,29 +1,107 @@
...
         sys.exit(-1)
 
+    hdfs_connection.init_hdfs(config.HDFS_HTTP_URI)
+    if rm:
+        try:
+            utils.delete_dir(path.RECOMMENDATION_PARENT_DIR)
+            utils.delete_dir(path.CHECKPOINT_DIR)
+            logging.info('Successfully deleted directories.')
+        except HdfsError as err:
+            logging.error('{}: Some/all directories are non-empty. Try "--f" to delete recursively.'.format(
```
mayhem: Great error message!

manage.py
```diff
@@ -1,29 +1,107 @@
...
+                type(err).__name__))
+            logging.warning('Deleting directory recursively will delete all the recommendation data.')
+            sys.exit(-1)
+
+    if f:
+        utils.delete_dir(path.RECOMMENDATION_PARENT_DIR, recursive=True)
+        utils.delete_dir(path.CHECKPOINT_DIR, recursive=True)
+        logging.info('Successfully deleted directories recursively.')
+
+    if create_dir:
```
mayhem: What happens if these dirs already exist?

manage.py
```diff
@@ -1,29 +1,107 @@
...
+
+        logging.info('Creating directory to store candidate sets...')
+        utils.create_dir(path.CANDIDATE_SET_DIR)
+
+        logging.info('Creating directory to store RDD checkpoints...')
+        utils.create_dir(path.CHECKPOINT_DIR)
+
+        print('Done!')
+
+@cli.command(name='dataframe')
```
mayhem: Each of these 5 invocations below all do exception catching on their own, yes?

manage.py
vansika: https://hdfscli.readthedocs.io/en/latest/api.html#hdfs.client.Client.makedirs

If the dirs already exist, nothing happens. And if they don't, dirs with specified path are created. I have tested it in both the cases. If you want a check, we can first call status func, if the dirs already exist, we will get HDFS error, we can catch it and tell user that they exist, otherwise go ahead and create them

mayhem: I think that is fine for now -- if we get into trouble later, we can improve it.

manage.py
vansika: Yes

manage.py
vansika: <3

PR #46, opened 2019-08-16
Import paths from path.py in create_dataframes

vansika:
use recently created path.py to import required paths

### DIFF
--- a/listenbrainz_spark/recommendations/create_dataframes.py
+++ b/listenbrainz_spark/recommendations/create_dataframes.py
@@ -7,8 +7,10 @@ from datetime import datetime
 from py4j.protocol import Py4JJavaError
 
 import listenbrainz_spark
+from listenbrainz_spark import path
 from listenbrainz_spark import stats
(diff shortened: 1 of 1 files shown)


listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -17,6 +19,8 @@
 
 # dataframe html is generated when set to true
 SAVE_DATAFRAME_HTML = False
+# os.path.join ignores everything before a '/'
+HDFS_PATH = config.HDFS_CLUSTER_URI + path.DATAFRAME_DIR + '/'
```
vansika: @mayhem  I really need your comments on this.

mayhem: Yes, os.path.join is to join filenames consistently across various filesystems (windows, linux, mac). However, given that we're joining paths for HDFS and that this isn't really supported in os.path, I'm ok with this.

PR #48, opened 2019-08-16
Remove zip created on every run of spark-submit

vansika:
spark-submit's --py-file param takes python files in .zip, .egg etc format. We are using .zip format to create a zip of listenbrainz_spark folder. On every run, spark-submit creates a new zip (to include modifications) and save it in the location from where spark-submit has been called. 
After a script has been executed, this zip file is of no use and hence should be deleted.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -106,6 +106,11 @@ def user():
     from listenbrainz_spark.stats import user
     user.main()
 
+@cli.resultcallback()
+def remove_zip(result, **kwargs):
+    """ Remove zip created by spark-submit.
+    """
+    os.remove(os.path.join('/', 'rec', 'listenbrainz_spark.zip'))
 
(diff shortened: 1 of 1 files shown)


paramsingh:
Hey, this is a question, not something that needs to be done in this pulll request, lekin do we have some documentation on how to actually run the commands in manage.py?

vansika:
@paramsingh not yet. But I think it is a great idea. How about using hacking.md for the same?

paramsingh:
It'd be great if you added that, I've been a little out of touch with this, so I'm not really sure about the commands we can run.

vansika:
Yup. I will open a PR for that soon :)

paramsingh:
Alright, can you open a ticket in the meanwhile?

PR #49, opened 2019-08-21
Add a new docker image to set up cron for statistics

paramsingh:
The container would be running cron jobs from the cluster. this seems like an okay plan for now while we're still on the hetzner vms and if we go to permanent hetzner machines, it should be good too. 

## Open questions

* How do we handle private configs in the vms? the vms do not have consul. 
* Is it possible right now to connect to the rabbitmq instance on our main machines from the VMs? If so, what's the IP address / host and port?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,3 +49,14 @@ COPY . /rec
 
 FROM metabrainz-spark-base as metabrainz-spark-dev
 COPY . /rec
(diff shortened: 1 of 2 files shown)


paramsingh:
Decided that we won't run cron here, and the listenbrainz-server would send us requests.

PR #50, opened 2019-08-27
Create schema to store dataframe metadata and connect it with model metadata schema

vansika:
The basic approach is to invoke _dataframe_metadata_ with _create_dataframes.py_ At this point, we will generate a  _model_id_ ,store it in _dataframe_metadata_ and _model_metadata_ and set _updated_ to False.

Later when _model_metadata_ is invoked with _train_models,_ the _model_id_ with _updated_ = False will be used as best model id, other fields would be updated and _updated_  will be set to True


vansika:
@alastair @mayhem @paramsingh   Please review. (since only repo owners and collaborators can request a review, requested in comment)

PR #51, opened 2019-09-01
Include all file and dir paths in a single file

vansika:


listenbrainz_spark/path.py
```diff
@@ -1,12 +1,34 @@
 import os
+import config
+
+# Parent directory containing listens and data driven from ListenBrainz.
+LISTENBRAINZ_DATA_DIRECTORY = os.path.join('/', 'data', 'listenbrainz')
```
paramsingh: Are there places in the codebase we should change to use this? I think we should do that in this PR as well.

listenbrainz_spark/path.py
```diff
@@ -1,12 +1,34 @@
...
+# Directory containing similar artist relation.
+# (This is a temporary path till incremental dumps for similar artists are prepared)
+SIMILAR_ARTIST_DIR = os.path.join('/', 'data', 'listenbrainz', 'similar_artists')
```
mayhem: As param pointed out, we should use the LISTENBRAINZ_DATA_DIRECTORY for anything that goes under /data/listenbrainz so that if we need to move it, we can change one value only.

listenbrainz_spark/path.py
```diff
@@ -1,12 +1,34 @@
...
+
 
 # Parent directory containing data used in and generated by recommendation engine.
 RECOMMENDATION_PARENT_DIR = os.path.join('/', 'recommendation')
 # Directory containing dataframes to be used in generating recommendations (not necessarily).
 DATAFRAME_DIR = os.path.join('/', 'recommendation', 'dataframe')
```
mayhem: And here DATAFRAME_DIR should use RECOMMENDATION_PARENT_DIR..

listenbrainz_spark/path.py
```diff
@@ -1,12 +1,40 @@
...
 # Directory containing candidate sets to be used for generating recommendations.
-CANDIDATE_SET_DIR = os.path.join('/', 'recommendation', 'candidate_set')
+CANDIDATE_SET_DIR = os.path.join(RECOMMENDATION_PARENT_DIR, 'candidate_set')
 # Directory containing RDD checkpoints to break lineage while using iterative algorithms.
 CHECKPOINT_DIR = os.path.join('/', 'checkpoint')
+# Directory to save best models
+DATA_DIR = os.path.join(MODEL_DIR, 'data')
+
+# Absolute path to dataframes used in processing raw data/listens.
+USERS_DATAFRAME_PATH = config.HDFS_CLUSTER_URI + DATAFRAME_DIR + '/' + 'users_df.parquet'
```
paramsingh: any particular reason for us to not use os.path.join here?

paramsingh: and in the lines that follow

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -256,8 +249,8 @@ def main():
         sys.exit(-1)
     logging.info('Files fetched from HDFS and dataframes registered in {}s'.format('{:.2f}'.format(time() - ti)))
 
-    path = os.path.join(os.path.dirname(os.path.abspath(__file__)),'recommendation-metadata.json')
-    with open(path) as f:
+    _path = os.path.join(os.path.dirname(os.path.abspath(__file__)),'recommendation-metadata.json')
```
paramsingh: variables starting with `_` are generally private variables in classes, just name it metadata_file_path?

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -228,14 +224,13 @@ def main():
         logging.error('{}\n{}\nAborting...'.format(str(err), err.java_exception))
         sys.exit(-1)
 
-    path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'recommendation-metadata.json')
-    with open(path, 'r') as f:
+    _path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'recommendation-metadata.json')
```
paramsingh: this could be called `metadata_file_path` too

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -228,14 +224,13 @@ def main():
...
+    with open(_path, 'r') as f:
         recommendation_metadata = json.load(f)
         best_model_id = recommendation_metadata['best_model_id']
         user_names = recommendation_metadata['user_name']
 
-    best_model_path = os.path.join('/', 'data', 'listenbrainz', 'recommendation-engine', 'best-model', '{}' \
-        .format(best_model_id))
+    best_model_path = path.DATA_DIR + '/' + best_model_id
```
paramsingh: os.path.join?

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -214,9 +211,9 @@ def main():
             models_training_time)
 
     # Save best model id to a JSON file
-    path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'recommendation-metadata.json')
-    with open(path, 'r') as f:
+    _path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'recommendation-metadata.json')
```
paramsingh: same here.

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -200,9 +198,8 @@ def main():
     logging.info('Saving model...')
     t0 = time()
     try:
-        path = os.path.join('/', 'data', 'listenbrainz', 'recommendation-engine', 'best-model', '{}' \
-        .format(best_model_metadata['model_id']))
-        model.model.save(listenbrainz_spark.context, config.HDFS_CLUSTER_URI + path)
+        _path = path.DATA_DIR + '/' + best_model_metadata['model_id']
+        model.model.save(listenbrainz_spark.context, config.HDFS_CLUSTER_URI + _path)
```
paramsingh: os.path.join in the line here and the one above?

listenbrainz_spark/path.py
vansika: https://stackoverflow.com/questions/1945920/why-doesnt-os-path-join-work-in-this-case

listenbrainz_spark/recommendations/train_models.py
mayhem: os.path.join doesn't work with HDFS paths.

vansika: The problem here is not an HDFS path mainly. For instance, I have a dir path as:
` DATA_DIR = os.path.join('/', 'listenbrainz', 'data') `
DATA_DIR = '/data/listenbrainz'
Now, when I want to access a file from DATA_DIR I would have to specify the full path which is:
 'hdfs://hadoop-master:9000/data/listenbrainz'

HDFS_CLUSTER_URI =  'hdfs://hadoop-master:9000'
`FINAL_PATH = os.path.join(HDFS_CLUSTER_URI, DATA_DIR)` 
FINAL_PATH  = '/data/listenbrainz' which is defintely not what we want since os.path.join ignores everything before a '/'
That is why we have used '+'. Note that FINAL_PATH is an url and not exactly a path.

We can do something like this if os.path.join must be used
`os.path.join('hdfs://hadoop-master:9000', 'data', 'listenbrainz')`
but there are places where we just want DATA_DIR and not the full URL. In order to maintain flexibility I have kept DIRs and URLs diff.

PR #53, opened 2019-09-17
Install Python 3.6 ourselves instead of using old jessie python

paramsingh:
Python 3.4 is not supported so we shouldn't use it. Debian jessie, which is what the oraclejdk image is based on, doesn't have python higher than 3.4 in its repositories.

Not using python 3.7 because it needs openssl 1.0.2 which is
again not present in jessie (*sigh*)

We really need to think about whether we should just start using base image (https://github.com/metabrainz/docker-python) and install oracle jdk ourselves (https://launchpad.net/~webupd8team/+archive/ubuntu/java).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,132 @@
 FROM airdock/oraclejdk:1.8 as metabrainz-spark-base
 
-RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
+# Compile and install specific version of Python
+# The jdk image comes with jessie which has python 3.4 which
+# is not supported anymore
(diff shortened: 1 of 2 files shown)


paramsingh:
@vansika this is running on the cluster now, it'd be great if you could test once that the recommendation scripts are working as expected.

Dockerfile
```diff
@@ -1,22 +1,146 @@
 FROM airdock/oraclejdk:1.8 as metabrainz-spark-base
 
-RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
+# Compile and install specific version of Python
+# The jdk image comes with jessie which has python 3.4 which
+# is not supported anymore
```
alastair: might want to additionally say here that we're installing 3.6 because of openssl, especially since the following line says 3.7

alastair:
looksgood

PR #55, opened 2019-09-26
Add checkpoint to break and save RDD lineage during iterative process

vansika:
During iterative calls (usually when iterations > 10 ) Spark forgets RDD lineage. It is recommended to add checkpoints to break the long lineage of an RDD, save the breakpoint as a new base so that when the RDD is computed again in the next iteration the distance between base and the result is shorter.

### DIFF
--- a/listenbrainz_spark/recommendations/train_models.py
+++ b/listenbrainz_spark/recommendations/train_models.py
@@ -15,7 +15,8 @@ import listenbrainz_spark
 from listenbrainz_spark import hdfs_connection
 from listenbrainz_spark import config, utils, path
 from listenbrainz_spark.recommendations.utils import save_html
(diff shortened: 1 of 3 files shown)


listenbrainz_spark/recommendations/train_models.py
```diff
@@ -154,6 +156,9 @@ def main():
         current_app.logger.error(str(err), exc_info=True)
         sys.exit(-1)
 
+    # Add checkpoint dir to break and save RDD lineage.
+    listenbrainz_spark.context.setCheckpointDir(config.HDFS_CLUSTER_URI + path.CHECKPOINT_DIR)
```
mayhem: Does Spark create the dir if it doesn't exist? If not, what creates it?

vansika: Yes, Spark creates the directory if it does not exist.

PR #56, opened 2019-09-27
Remove recommendation engine's dependency on dir initialisation

vansika:
As suggested by @alastair we should create directories with the script which need them. It may happen that we run a three-hour job and later realise that the required directory was not created. 

The recommendation engine files will be run in a specific order which will ensure that all the directories will be created before we need them, we will probably have a cron job ensuring that.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -4,74 +4,13 @@ import click
 import logging
 
 import listenbrainz_spark
-from listenbrainz_spark import path
 from listenbrainz_spark import utils
-from listenbrainz_spark import config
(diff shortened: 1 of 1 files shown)


mayhem:
Are  you planning on adding the needed bits to create the dirs as the scripts go? I think that should be part of this PR.

vansika:
@mayhem The directories will be created as and when needed with respective scripts, and their order can be looked upon in tests.

paramsingh:

PR #58, opened 2019-10-14
Keep track of view names to use same dataframe in different modules.

vansika:
right now, view names are hard-coded which makes it difficult to debug and handle code. View.py properly formats view names and makes it easy to register the same dataframe between different modules.


vansika:
Since we are gradually shifting to pyspark SQL API, this PR is no longer required.

PR #60, opened 2019-10-22
Use mbid->msid mapping to get mbid for every listen

vansika:
As discussed we will use mbids to generate recommendations instead of msids.


listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -68,13 +67,14 @@ def get_listens_for_training_model_window(metadata):
...
+            A dataframe with columns as:
                 [
-                    artist_mbids, artist_msid, artist_name, listened_at, recording_mbid,
-                    recording_msid, release_mbid, release_msid, release_name, tags, track_name, user_name
+                    artist_msid, artist_name, listened_at, recording_msid, release_mbid,
+                    release_msid, release_name, tags, track_name, user_name
                 ]
     """
     to_date = datetime.utcnow()
+    to_date = stats.adjust_days(to_date, 5000)
```
mayhem: Is 5000 supposed to be a config value as the comment says?

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -89,7 +89,34 @@ def get_listens_for_training_model_window(metadata):
     except FileNotFetchedException as err:
         current_app.logger.error(str(err), exc_info=True)
         sys.exit(-1)
-    return training_df
+    return utils.get_listens_without_artist_and_recording_mbids(training_df)
+
+def get_mapped_artist_and_recording_msids(partial_listens_df, artist_mapping_df, recording_mapping_df):
+    """ Map recording msid->mbid and artis msid->mbids so that every listen has an mbid.
```
mayhem: artis -> artist

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -89,7 +89,34 @@ def get_listens_for_training_model_window(metadata):
     except FileNotFetchedException as err:
         current_app.logger.error(str(err), exc_info=True)
         sys.exit(-1)
-    return training_df
+    return utils.get_listens_without_artist_and_recording_mbids(training_df)
+
+def get_mapped_artist_and_recording_msids(partial_listens_df, artist_mapping_df, recording_mapping_df):
```
mayhem: This fetches MBIDs, no? Would a better name for this function be get_mapped_artist_and_recording_mbids ?

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -103,29 +130,24 @@ def main():
...
 
     current_app.logger.info('Preparing users data and saving to HDFS...')
     t0 = time()
-    try:
-        users_df = sql.prepare_user_data(table)
-    except SQLException as err:
-        current_app.logger.error(str(err), exc_info=True)
-        sys.exit(-1)
+    # We use window function to give rank to distinct user_names
+    # Not that if user_names are not distinct rank would repeat and give unexpected results.
```
mayhem: note?

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -103,29 +130,24 @@ def main():
...
+    # users_df = | user_name| user_id|
+    user_window = Window.orderBy('user_name')
+    users_df = complete_listens_df.select('user_name').distinct().withColumn('user_id', rank().over(user_window))
```
mayhem: I wonder if this should be in a separate function for cleanness sake?

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -152,27 +172,17 @@ def main():
     recordings_df_time = '{:.2f}'.format((time() - t0) / 60)
 
     current_app.logger.info('Preparing listen data dump and playcounts, saving playcounts to HDFS...')
-    t0 = time()
-    try:
-        listens_df = sql.prepare_listen_data(table)
```
mayhem: What happened to this function? Is it still in use? Not sure I like that all the logic has moved to the top level function...

listenbrainz_spark/recommendations/create_dataframes.py
vansika: I was experimenting with something, forgot to remove. Thanks :)

listenbrainz_spark/recommendations/create_dataframes.py
vansika: oops :(

listenbrainz_spark/recommendations/create_dataframes.py
vansika: definitely :)

listenbrainz_spark/recommendations/create_dataframes.py
vansika: `listens_df = complete_listens_df.select('recording_mbid', 'user_name')`
This does the same thing as the above function. As we discussed, we will slowly deprecate the usage of queries and use API instead.  
What do you mean by top-level? And what specifically you don't like about it...

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -152,27 +172,17 @@ def main():
...
-        current_app.logger.error(str(err), exc_info=True)
-        sys.exit(-1)
+    # playcounts_df: | recording_id| user_id| count|
+    # listens_df is joined with users_df on user_name which gives us [ recording_mbid', 'user_name', user_id].
+    # The output is then joined with recording_df on recording_mbid which results into ['recording_mbid', 'recording_id', 'user_name', 'user_id'],
+    # The final step is groupBy which create groups on user_id and recording_id and count the number of recording_ids.
+    # The final dataframe tells us about the number of times a user has listend to a particular track for all users.
+    playcounts_df = listens_df.join(users_df, 'user_name', 'inner') \
+            .join(recordings_df, 'recording_mbid', 'inner') \
+            .groupBy('user_id', 'recording_id').agg(func.sum('recording_id').alias('count'))
```
vansika: should be func.count....

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -11,77 +11,122 @@
...
     """
-    top_artists = [row.artist_name for row in top_artists_df.collect()]
+    top_artists = [row.mb_artist_credit_id for row in top_artists_df.collect()]
 
-    if len(top_artists) == 1:
-        # Handle tuple with single entity
-        similar_artists_df = sql.get_similar_artists_with_limit(tuple(top_artists[0]))
-    else:
-        similar_artists_df = sql.get_similar_artists_with_limit(tuple(top_artists))
+    similar_artists_df = artists_relation_df.select(col('id_0').alias('top_artist_credit_id'), col('id_1') \
```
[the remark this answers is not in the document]
vansika: It makes sense to line up all the sub calls vertically but .alias() should stick with col() since it is changing name of that particular column.

vansika:
Would push tests soon, then it is ready to go :)

PR #61, opened 2019-11-09
scripts to start/stop request-consumer container

mayhem:
As discussed here:

https://www.dropbox.com/scl/fi/9tct95wsp5emkl1dp1pdu/%20ListenBrainz%20Statistics%20Todo.paper?dl=0&rlkey=048b4w9wwlflbd7panmpq1pg3

### DIFF
--- a/docker/stop-master-service.sh
+++ b/docker/stop-master-service.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
 
 docker service rm spark-master hadoop-master
+
+# stop the request consumer
+./stop-request-consumer-container.sh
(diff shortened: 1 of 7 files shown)


docker/stop-master-service.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 
 docker service rm spark-master hadoop-master
+
+# stop the request consumer
+./stop-request-consumer-container.sh
```
paramsingh: Should we do this before removing the services as it connects to the services?

PR #62, opened 2019-11-09
Add query map and module for queries

paramsingh:
First cut to see if I'm going in the right direction.


paramsingh:
Added the `listenbrainz_spark/stats/user` module which can have a file per query here in [request_queries.json](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/spark/request_queries.json), does that make sense?

PR #63, opened 2019-11-09
Add request consumer 

paramsingh:
This code adds a request consumer which listens for requests from lemmy in rabbitmq, takes them, gets a response and pushes it back into a different response queue.

Very untested code right now.

### DIFF
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -29,28 +29,28 @@ from flask import current_app
 
 class RequestConsumer:
 
-	def get_response(self, request):
-		query = request['query']
-		params = request['params']
+    def get_response(self, request):
+        query = request['query']
+        params = request['params']
 
-		try:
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -0,0 +1,114 @@
...
+            try:
+                self.response_channel.basic_publish(
+                    exchange=current_app.config['SPARK_RESULT_EXCHANGE'],
+                    routing_key='',
+                    body=ujson.dumps(response),
+                    properties=pika.BasicProperties(delivery_mode = 2,),
+                )
+                break
+            except pika.exceptions.ConnectionClosed:
+
```
mayhem: missing something here? pass?

mayhem: sleep() ?

paramsingh: that was embarrassing.

PR #65, opened 2019-11-10
Working request consumer

paramsingh:
There were a lot of bugs to fix. missing imports, uncaught exceptions and other stuff

I also made it a bit more production ready, by reopening closed channels and other such stuff.

This fixes the pipeline, but the queries need to be fixed. It OOMed out on me a lot of times when we were calculating artists, recordings and releases together. The rabbitmq messages we're sending in responses are definitely much much bigger than ideal. Definitely stuff to think about. But hey, we have a working pipeline! :tada: :tada: :tada:


listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -54,26 +58,48 @@ def get_response(self, request):
...
+            current_app.logger.info("Pushing to result queue...")
+            self.push_to_result_queue(result)
+            current_app.logger.info("Done!")
+        while True:
+            try:
+                self.request_channel.basic_ack(delivery_tag=method.delivery_tag)
+                break
+            except pika.exceptions.ChannelClosed:
+                self.request_channel = self.rabbitmq.channel()
+                self.request_channel.exchange_declare(exchange=current_app.config['SPARK_REQUEST_EXCHANGE'], exchange_type='fanout')
```
mayhem: What is going on here with the indent?

paramsingh: what indent problem?

mayhem: oh, I see -- if the window is too small it looks like the indent is off.

PR #66, opened 2019-11-13
Schema to store model and dataframe metadata

vansika:
Till now we were using a JSON file to keep track of recently trained model so that recommend.py can use that model string to load the model from HDFS.

This PR uses dataframe for storing the model id of the recently trained model. Note that right now the idea of data frame with only one row may seem absurd but in the near future, we will want to have multiple models saved in HDFS and use them in some fashion.


listenbrainz_spark/recommendations/train_models.py
```diff
@@ -105,16 +194,136 @@ def train(training_data, validation_data, num_validation, ranks, lambdas, iterat
...
+
+def delete_model(model_id):
+    """ Delete a model.
+
+        Args:
+            model_id (str): Model identification string of model to be deleted.
+    """
+    utils.delete_dir(path.DATA_DIR + model_id, recursive=True)
+
+def save_model_delete_metadata(model_id, timestamp):
```
mayhem: what? this is very confusing and the docstring doesn't really help at all.

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -50,6 +70,75 @@ def compute_rmse(model, data, n):
       .values()
     return sqrt(predictionsAndRatings.map(lambda x: (x[0] - x[1]) ** 2).reduce(add) / float(n))
 
+def generate_best_model_id():
+    """ Generate best model id.
+    """
+    return '{}-{}'.format(config.MODEL_ID_PREFIX, uuid.uuid4())
+
+def update_dataframe_metadata(best_model_id, new_best_model_id, dataframe_metadata_df):
+    """ Insert a row in dataframe_metadata dataframe.
```
mayhem: The function name says one thing, but the docstring say something different. One of these two is incorrect.

mayhem: Another thing to consider is that if a function may do different operations depending on the state of the DB, you can also name functions like "insert_or_update_dataframe_metadata", which would insert new metadata or update the metadata if it already exists.

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -13,21 +13,41 @@
...
+import pyspark.sql.functions as f
 from pyspark.mllib.recommendation import ALS, Rating
 
-Model = namedtuple('Model', 'model error rank lmbda iteration model_id training_time rmse_time')
+Model = namedtuple('Model', 'model validation_rmse rank lmbda iteration model_id training_time rmse_time')
 
 # training HTML is generated if set to true
 SAVE_TRAINING_HTML = True
 
+def best_model_id_exists(best_model_id, model_metadata_df):
```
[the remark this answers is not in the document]
vansika: Sounds cool. But just one concern.....That model may have been deleted from HDFS. We just want to know if the model id generated by create_datafarmes has been used by once or not. Do you still think that we should change the name to fnd_best_model?

listenbrainz_spark/recommendations/train_models.py
vansika: I agree. Thanks for pointing out :)

vansika: how about save_deleted_model_ids?

listenbrainz_spark/recommendations/train_models.py
vansika: I understand. But just FYI dataframes in Spark are immutable, so technically you cannot add some data to existing rows. Now, metaphorically here we are updating the dataframe by adding a new row. How come?

So, imagine train_models is generating model_id for itself since model_ids generated by create_dataframes have been exhausted. Nice. But do you see the catch? We don't have an entry in create_dataframes of *this* model id. And I hope you understand that the dataframe metadata of this model_id would be exactly same as the recently added row of dataframe metadata, so we *update* dataframe_metadata table (technically insert a row ). 

This was just to make the algo clear. So would you like to suggest a name? I was thinking of insert_dataframe_metadata?

vansika: Also, to be more technical, you cannot even insert a row to existing dataframe. In the back it will create a new dataframe, copy all the rows to new dataframe along with the row you wish to add.

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -19,10 +19,10 @@
 # dataframe html is generated when set to true
 SAVE_DATAFRAME_HTML = False
 
-def generate_best_model_id(metadata):
-    """ Generate best model id.
+def generate_dataframe_id(metadata):
```
mayhem: ahhhhh, much better.

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -50,6 +52,27 @@ def compute_rmse(model, data, n):
       .values()
     return sqrt(predictionsAndRatings.map(lambda x: (x[0] - x[1]) ** 2).reduce(add) / float(n))
 
+def generate_model_id():
+    """ Generate best model id.
+    """
+    return '{}-{}'.format(config.MODEL_ID_PREFIX, uuid.uuid4())
+
+def get_dataframe_id(dataframe_metadata_df, best_model_metadata):
+    """ Get dataframe id of datafsets on which model is trained from dataframe metadata.
```
mayhem: datafsets

mayhem: And there are still a number of "best_model_metadata" references in the code/docstrings, which I find confusing. Can you please clean those up?

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -50,6 +52,27 @@ def compute_rmse(model, data, n):
       .values()
     return sqrt(predictionsAndRatings.map(lambda x: (x[0] - x[1]) ** 2).reduce(add) / float(n))
 
+def generate_model_id():
+    """ Generate best model id.
```
mayhem: best

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -105,16 +129,96 @@ def train(training_data, validation_data, num_validation, ranks, lambdas, iterat
...
                 model_id=model_id, training_time=mt, rmse_time=vt)
 
-    best_model_metadata = {'error': '{:.2f}'.format(best_model.error), 'rank': best_model.rank, 'lmbda':
+    best_model_metadata = {'validation_rmse': best_model.validation_rmse, 'rank': best_model.rank, 'lmbda':
             best_model.lmbda, 'iteration': best_model.iteration, 'model_id': best_model.model_id, 'training_time':
-                best_model.training_time, 'rmse_time': best_model.rmse_time}
+                best_model.training_time, 'rmse_time': best_model.rmse_time, 'alpha': alpha}
     return best_model, model_metadata, best_model_metadata
 
+def delete_model():
```
mayhem: This deletes all the models??

mayhem: OK, as per our discussion in IRC, this is fine for now.

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -183,33 +292,31 @@ def main():
     current_app.logger.info('Training models...')
     t0 = time()
 
-    model, model_metadata, best_model_metadata = train(training_data, validation_data, num_validation, config.RANKS,
-        config.LAMBDAS, config.ITERATIONS)
+    model, model_metadata, best_model_metadata = train(training_data, validation_data, num_validation,
```
mayhem: I guess here best_model is actually quite accurate, no?

listenbrainz_spark/recommendations/train_models.py
vansika: > And there are still a number of "best_model_metadata" references in the code/docstrings, which I find confusing.

which references?  I guess those references are needed. Can you please point some so that I can understand better?

listenbrainz_spark/recommendations/train_models.py
vansika: This function is generating model ids for many models, so I feel that the docstring should be corrected.

listenbrainz_spark/recommendations/train_models.py
vansika: sure

vansika:
Sure :)

paramsingh:

PR #67, opened 2019-11-13
Setup tests

paramsingh:
* Add a base test class
* Add test.sh
* Add jenkins builds

### DIFF
--- a/listenbrainz_spark/tests/test_utils.py
+++ b/listenbrainz_spark/tests/test_utils.py
@@ -1,0 +1,14 @@
+import unittest
+
+from listenbrainz_spark import utils
+from listenbrainz_spark.tests import SparkTestCase
+
+class UtilsTestCase(SparkTestCase):
+
+    def test_create_dataframe(self):
+        df = utils.create_dataframe({'column1': 1, 'column2': 2}, schema=None)
+        self.assertEqual(df.count(), 1)
+        path = 'hdfs://hadoop-master:9000/test_df.parquet'
+        utils.save_parquet(df, path)
+        received_df = utils.read_files_from_HDFS(path)
+        self.assertEqual(received_df.count(), 1)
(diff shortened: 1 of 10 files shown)


paramsingh:
@brainzbot test this please

paramsingh:
@brainzbot test this please

listenbrainz_spark/tests/test_utils.py
```diff
@@ -0,0 +1,14 @@
...
+from listenbrainz_spark.tests import SparkTestCase
+
+class UtilsTestCase(SparkTestCase):
+
+    def test_create_dataframe(self):
+        df = utils.create_dataframe({'column1': 1, 'column2': 2}, schema=None)
+        self.assertEqual(df.count(), 1)
+        path = 'hdfs://hadoop-master:9000/test_df.parquet'
+        utils.save_parquet(df, path)
+        received_df = utils.read_files_from_HDFS(path)
```
vansika: Since this test function tests utils.create_dataframe we should not use utils.read_files_from_HDFS in here, we will have a separate test for that which will be enough to test the function. Using functions other than the function being tested would lead to many changes in the test script if we make modifications in the codebase. Ideally, we should use 
`listenbrainz_spark.sql_context.read.parquet(path)`

I will take care of this while writing tests for utils. I thought it will be good to mention :)

paramsingh: I don't really agree here. There's no rule that a test needs to only call the function mentioned in it's name. In this case, `read_files_from_HDFS` provides a useful abstraction over the SQL context so using it like you'd use it in actual code makes sense. 

Also, in general, tests should be written in such a way that they capture what the code is _really_ trying to do. For example, no matter how the implementation of the readfilesfromhdfs function changes, it'll always read and return a file from hdfs. Hence, if you have to modify the usage of the function in this test, you should really think about what you're doing, because the function is probably used in the same way in lots of places in real code.

paramsingh: The name of this test should be changed though.

vansika: I did not think from this angle. Thank you :)

PR #68, opened 2019-11-26
Fix test.sh by removing the datanode volume

paramsingh:
This allows us to format namenodes and datanodes will always
be ok.

cc @vansika, please test this. I've tested it on leader and test.sh works for me.

### DIFF
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -1,7 +1,6 @@
 version: "3.4"
 
 volumes:
-    datanode:
     namenode:
 
 services:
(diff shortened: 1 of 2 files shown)


vansika:
This solution works but we need something else too.
What I have noticed is that before every run of test.sh, I need to run the docker command to *format name node*. This *format-with-every-run* happened after I uploaded test listens to HDFS. What I think the reason is:
Namenode stores a file called fsimage to keep track of data blocks in datanode and before executing any job waits for datanodes to send count and other metadata about the data blocks. We are not making volume for datanode but for namenode so namenode waits to match the information in its fsimage (which retains becuse of volume) but datanode has no data blocks and it leads to weird errors and namenode going into safe mode.

Can we add the format command permanently in test.sh to format namenode with every run?

paramsingh:
Done! Please take a look.

PR #69, opened 2019-11-28
Generate and upload test listens to HDFS 

vansika:
With every run of test.sh listens of the timestamp of current and previous month would be added to JSON files in listenbrainz_spark/testdata and uploaded to HDFS to perform tests.


vansika:
@paramsingh  Oh yes. So There are two JSON files in listenbrainz_spark/testdata/test_listens which contain listens except the *listened_at* key since every time the test.sh runs we want to upload listens to HDFS that are of previous and the present month to check functions like these https://github.com/metabrainz/listenbrainz-labs/blob/master/listenbrainz_spark/sql/candidate_sets_queries.py#L20

So with this PR, we iterate over the listens and assign timestamp by starting from the first day of month to the last ( and loop again) for two months. 

The number of listens is kept such so that we can train the model as well ( which requires an appreciable number of listens)

pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
-testpaths = listenbrainz_spark
+testpaths = tests listenbrainz_spark/tests listenbrainz_spark/recommendations
```
paramsingh: All code should be inside the `listenbrainz_spark` module because it makes imports etc easier. I think you should just move the file you have in `tests` to `listenbrainz_spark/tests`.

For context, we've been burned by not having code in a single module leading to us having to add stuff to `sys.path` and needing major refactors.

tests/__init__.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import uuid
+import listenbrainz_spark
+
+from listenbrainz_spark import hdfs_connection
+
+HDFS_HTTP_URI = 'http://hadoop-master:9870' # the URI of the http webclient for HDFS
```
paramsingh: this should come from config.

tests/test_generate_listens.py
```diff
@@ -0,0 +1,94 @@
...
+from dateutil.relativedelta import relativedelta
+
+import listenbrainz_spark
+from tests import SparkTestCase
+from listenbrainz_spark import utils
+from listenbrainz_spark import hdfs_connection
+
+from pyspark.sql.types import StructField, StructType, ArrayType, StringType, TimestampType, FloatType, IntegerType, BooleanType
+
+HDFS_CLUSTER_URI = 'hdfs://hadoop-master:9000' # the URI to be used with Spark
```
paramsingh: This should come from config.

tests/test_generate_listens.py
```diff
@@ -0,0 +1,94 @@
...
+TESTDATA_PATH = '/rec/listenbrainz_spark/testdata/test_listens/' # local path to test listens
+LISTENS_PATH = '/test/listens/' # HDFS path to listens
+
+END_DATE = datetime.utcnow().replace(microsecond=0).replace(day=1) # For listens of the current month
+START_DATE = (END_DATE + relativedelta(months=-1)) # For listens of the previous month.
+
+class GenerateTestData(SparkTestCase):
```
paramsingh: This class is not a `TestCase` because we're not testing any real code here. If you need listens for a particular test or a group of tests, you should add them to the testcase class for those listens. I think keeping tests pure in the sense that they have no side-effects (keep hdfs in the same state that it was before the tests ran) is a thing to aim for. 

In listenbrainz, for database test cases, we add the specific data in the `setUp` function of the test cases and delete the data afterwards. Does that make sense?

If you think that you'll have a large number of tests which will need this amount of data, then we should add this code to the setUp function of SparkTestCase, although I'm wary of doing that because it would make tests very slow. I think the best way forward here is to call `create_json_files` and `upload_listens_to_hdfs` in `setUpClass` in `CreateDataframeTestCase` in ` listenbrainz_spark/recommendations/tests/test_dataframe.py`.

tests/test_generate_listens.py
```diff
@@ -0,0 +1,94 @@
...
+        """ Invoke functions to create and upload test listens.
+        """
+        self.generate_timestamps_for_listens()
+        m1, m2 = START_DATE.strftime('%m'), END_DATE.strftime('%m')
+        self.upload_listens_to_hdfs(m1, START_DATE.strftime('%Y'))
+        self.upload_listens_to_hdfs(m2, END_DATE.strftime('%Y'))
+        # Remove the JSON files from local directory after listens have been uploaded to HDFS.
+        os.remove(TESTDATA_PATH + '{}.json'.format(m1))
+        os.remove(TESTDATA_PATH + '{}.json'.format(m2))
+
```
paramsingh: I'm gonna try to set up a linter for this but there's too many newlines at the end of this file and a few others. Please take a look! Sorry.

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,22 @@
...
+
+class CreateDataframesTestCases(SparkTestCase):
+
+    def test_users_dataframe(self):
+        training_df = utils.get_listens(START_DATE, END_DATE, HDFS_CLUSTER_URI + LISTENS_PATH)
+        m1 = training_df.select(f.min('listened_at').alias('listened_at')).take(1)[0].listened_at.month
+        m2 = training_df.select(f.max('listened_at').alias('listened_at')).take(1)[0].listened_at.month
+        self.assertEqual(m1, int(START_DATE.strftime('%m')))
+        self.assertEqual(m2, int(END_DATE.strftime('%m')))
+
```
paramsingh: newline fix needed.

pytest.ini
paramsingh: If you were doing this because of import problems while putting tests in directories other that `listenbrainz_spark/test`, I've fixed that problem in #72. Please take a look.

vansika:
This PR was the first step towards unit tests, gradually I learnt more about writing unit tests and now I feel these changes are no longer required.

PR #71, opened 2019-12-02
Unit tests for utils

vansika:
# Description
This PR adds unit tests to test listenbrainz_spark/utils.py

### DIFF
--- a/listenbrainz_spark/hdfs_connection.py
+++ b/listenbrainz_spark/hdfs_connection.py
@@ -5,8 +5,3 @@ client = None
 def init_hdfs(namenode_uri, user='root'):
     global client
     client = hdfs.InsecureClient(namenode_uri, user='root')
-
-def init_test_hdfs(namenode_uri, user='root'):
-    global client
-    client = hdfs.InsecureClient(namenode_uri, user='root')
-
--- a/listenbrainz_spark/hdfs_connection.py
+++ b/listenbrainz_spark/hdfs_connection.py
@@ -6,3 +6,7 @@ def init_hdfs(namenode_uri, user='root'):
     global client
     client = hdfs.InsecureClient(namenode_uri, user='root')
 
+def init_test_hdfs(namenode_uri, user='root'):
+    global client
+    client = hdfs.InsecureClient(namenode_uri, user='root')
+
--- a/listenbrainz_spark/tests/__init__.py
+++ b/listenbrainz_spark/tests/__init__.py
@@ -2,11 +2,16 @@ import unittest
 import uuid
 import listenbrainz_spark
 
+from listenbrainz_spark import hdfs_connection
+
+HDFS_HTTP_URI = 'http://hadoop-master:9870' # the URI of the http webclient for HDFS
+
 class SparkTestCase(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
         listenbrainz_spark.init_test_session('spark-test-run-{}'.format(str(uuid.uuid4())))
+        hdfs_connection.init_test_hdfs(HDFS_HTTP_URI)
 
     @classmethod
     def tearDownClass(cls):
--- a/listenbrainz_spark/tests/__init__.py
+++ b/listenbrainz_spark/tests/__init__.py
@@ -2,16 +2,14 @@ import unittest
 import uuid
 import listenbrainz_spark
 
-from listenbrainz_spark import hdfs_connection
-
-HDFS_HTTP_URI = 'http://hadoop-master:9870' # the URI of the http webclient for HDFS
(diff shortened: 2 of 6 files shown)


listenbrainz_spark/hdfs_connection.py
```diff
@@ -6,3 +6,7 @@ def init_hdfs(namenode_uri, user='root'):
     global client
     client = hdfs.InsecureClient(namenode_uri, user='root')
 
+def init_test_hdfs(namenode_uri, user='root'):
```
paramsingh: why is this needed? Shouldn't the normal hdfs connection function work? We use the same hostname in tests.

listenbrainz_spark/tests/__init__.py
```diff
@@ -2,11 +2,16 @@
 import uuid
 import listenbrainz_spark
 
+from listenbrainz_spark import hdfs_connection
+
+HDFS_HTTP_URI = 'http://hadoop-master:9870' # the URI of the http webclient for HDFS
```
paramsingh: This should come from the config...

listenbrainz_spark/tests/test_utils.py
```diff
@@ -1,14 +1,96 @@
...
+from datetime import datetime
 
-from listenbrainz_spark import utils
+import listenbrainz_spark
+from listenbrainz_spark import utils, hdfs_connection
 from listenbrainz_spark.tests import SparkTestCase
 
+from pyspark.sql import Row
+
+HDFS_CLUSTER_URI = 'hdfs://hadoop-master:9000'
```
paramsingh: This should come from config.

listenbrainz_spark/utils.py
```diff
@@ -186,18 +186,23 @@ def delete_dir(path, recursive=False):
...
     try:
-        hdfs_connection.client.delete(path, recursive=recursive)
+        deleted = hdfs_connection.client.delete(path, recursive=recursive)
+        if not deleted:
+            raise HDFSDirectoryNotDeletedException('', path)
+        return deleted
     except HdfsError as err:
         raise HDFSDirectoryNotDeletedException(str(err), path)
 
 def get_status(path):
```
paramsingh: This is very nitty, but I'm not sure if `get_status` is the best name for this function. `directory_exists` is much easier to understand.

listenbrainz_spark/tests/test_utils.py
```diff
@@ -1,14 +1,96 @@
...
+        df = utils.create_dataframe(Row(column1=1, column2=2), schema=None)
+        utils.append(df, hdfs_path)
+        new_df = utils.read_files_from_HDFS(hdfs_path)
+        self.assertEqual(new_df.count(), 1)
+
+        df = utils.create_dataframe(Row(column1=3, column2=4), schema=None)
+        utils.append(df, hdfs_path)
+        appended_df = utils.read_files_from_HDFS(hdfs_path)
+        self.assertEqual(appended_df.count(), 2)
+        # delete_dir dir in every test func so that using identical dirs by chance will not throw error.
```
paramsingh: this should be a part of the `tearDown` function so that it automatically gets called after each test and we don't need to duplicate code.

paramsingh: this should be a part of the `tearDown` function so that it automatically gets called after each test and we don't need to duplicate code.

listenbrainz_spark/utils.py
vansika: path_exists is better IMO, can be a file too.

listenbrainz_spark/tests/test_utils.py
vansika: We don't need to delete the path after every function call, utils.save_parquet overwrites anyway. I will remove all the calls but thanks for introducing me to the function, will be very useful in time :)

paramsingh: I still think making tests have no side effects makes sense, because some other test in some other file expects that directory to be empty.

vansika: I agree but some tests would not want the file to be empty like in recommendations where the results once saved would be used in between the test functions. I have left this part to be handled later so that we are sure when to use teardown, teardown_class, teardown_method or inheriting from a different parent class altogether.

paramsingh: OK, but in general, I would like tests to be independent of each other. If a test needs some file, it should create the file itself (or get it from its setup function). It should never expect another test to create the file because the order that tests run in is not guaranteed to be the same between test runs.

vansika: gotcha!

PR #73, opened 2019-12-04
Add __init__.py to recommendations module

paramsingh:
# Description

Make `listenbrainz_spark.recommendations` a python module.
# Problem

I noticed that our test coverage seemed higher than expected at 37%. (https://ci.metabrainz.org/job/ListenBrainz%20Labs/53/cobertura/) 

Then I realized that `recommendations` wasn't a python module (no `__init__.py`) meaning that py.test thought we only have 678 lines of code. Obivously we have more.

# Solution

Add `__init__.py` to `recommendations` so that it becomes a proper python module that can be imported from later.

# Action

 NA


paramsingh:
Coverage goes to 15% with ~1000 lines for this PR (https://ci.metabrainz.org/job/ListenBrainz%20Labs/54/cobertura/)

vansika:
<3

PR #74, opened 2019-12-13
Unit test listenbrainz_spark/recommendations

vansika:
# Description 
This PR contains tests for listenbrainz_spark/recommendations


listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+LISTENS_PATH = '/test/listens/' # HDFS path to listens
+MAPPING_PATH = '/test/mapping.parquet'
+MAPPED_LISTENS_PATH = '/test/mapped_listens.parquet'
+
+DATE = datetime.utcnow()
+
+class CreateDataframeTestclass(SparkTestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
paramsingh: Nice!

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+from listenbrainz_spark import schema, utils, config, path, hdfs_connection
+
+from pyspark.sql import Row
+
+# path used in between test functions of this class
+LISTENS_PATH = '/test/listens/' # HDFS path to listens
+MAPPING_PATH = '/test/mapping.parquet'
+MAPPED_LISTENS_PATH = '/test/mapped_listens.parquet'
+
+DATE = datetime.utcnow()
```
paramsingh: this variable being outside the test class and in constant style seems like an antipattern to me. I think we should just set it inside the test case as a member variable.

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+
+class CreateDataframeTestclass(SparkTestCase):
```
paramsingh: `CreateDataframeTestCase` for consistency

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
+import uuid
+import unittest
+from datetime import datetime
+
+import listenbrainz_spark
+from listenbrainz_spark.tests import SparkTestCase
+import listenbrainz_spark.recommendations.create_dataframes as dataframe
```
paramsingh: Why are we renaming the module here? It makes things a bit confusing to read.

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+    def upload_test_mapped_listens_to_HDFS(cls):
+        partial_listen_df = dataframe.get_listens_for_training_model_window(DATE, DATE, {}, LISTENS_PATH)
+        mapping_df = utils.read_files_from_HDFS(MAPPING_PATH)
+
+        mapped_df = dataframe.get_mapped_artist_and_recording_mbids(partial_listen_df, mapping_df)
+        utils.save_parquet(mapped_df, MAPPED_LISTENS_PATH)
+
+    def test_get_dates_to_train_data(self):
+        to_date, from_date = dataframe.get_dates_to_train_data()
+        self.assertLessEqual(from_date, to_date)
```
paramsingh: Might be a good idea to assert that the difference between the two dates is what we expect

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def test_get_listens_for_training_model_window(self):
+        metadata = {}
+        test_df = dataframe.get_listens_for_training_model_window(DATE, DATE, metadata, LISTENS_PATH)
+        self.assertEqual(metadata['to_date'], DATE)
+        self.assertEqual(metadata['from_date'], DATE)
+        self.assertNotIn(['artist_mbids', 'recording_mbid'], test_df.columns)
```
paramsingh: I'm not sure if this assertion does what you want it to do. It's checking that the list isn't in test_df.columns. Do we want to check that the artist_mbids and recording_mbid isn't in the columns instead? If so, we should use two separate assertNotIns

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,152 @@
...
+    def test_get_mapped_artist_and_recording_mbids(self):
+        partial_listen_df = dataframe.get_listens_for_training_model_window(DATE, DATE, {}, LISTENS_PATH)
+        mapping_df = utils.read_files_from_HDFS(MAPPING_PATH)
+
+        mapped_df = dataframe.get_mapped_artist_and_recording_mbids(partial_listen_df, mapping_df)
+        self.assertEqual(mapped_df.count(), 1)
+        complete_listen_col = ['artist_msid', 'artist_name', 'listened_at', 'recording_msid', 'release_mbid', 'release_msid',
+            'release_name', 'tags', 'track_name', 'user_name', 'mb_artist_credit_id', 'mb_artist_gids', 'mb_recording_gid',
+             'msb_artist_msid', 'msb_recording_msid']
+        self.assertEqual(complete_listen_col, mapped_df.columns)
```
paramsingh: Use assertListEqual here if we're comparing lists

listenbrainz_spark/recommendations/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
+import uuid
+import unittest
+
+import listenbrainz_spark
+from listenbrainz_spark.tests import SparkTestCase
+from listenbrainz_spark import utils, config, hdfs_connection
+import listenbrainz_spark.recommendations.train_models as model
+
+from pyspark.sql import Row
+from pyspark.sql.types import *
```
paramsingh: Wildcard imports are an antipattern that I'd like to avoid, can we just import what we need?

listenbrainz_spark/tests/test_utils.py
```diff
@@ -9,79 +9,71 @@
 from pyspark.sql import Row
 
 class UtilsTestCase(SparkTestCase):
+    # use path_ as prefix for all paths in this class.
+    path_ = '/test'
 
-    def test_append_dataframe(self):
-        path_ = 'test_df.parquet'
-        hdfs_path = os.path.join(config.HDFS_CLUSTER_URI, path_)
+    def setUp(self):
```
paramsingh: If this function isn't doing anything, you can just remove it.

listenbrainz_spark/utils.py
```diff
@@ -239,3 +235,21 @@ def path_exists(path):
...
+
+        Returns:
+            walk: a generator yeilding tuples (path, dirs, files).
+    """
+    try:
+        walk = hdfs_connection.client.walk(hdfs_path=path, depth=depth)
+        return walk
+    except HdfsError as err:
+        raise PathNotFoundException(str(err), path)
+
```
paramsingh: Extra whitespace, sorry to mention this everytime, but you should look into setting up your editor to remove these automatically

listenbrainz_spark/tests/test_utils.py
```diff
@@ -9,79 +9,71 @@
...
-        utils.create_dir(path_)
-
-        utils.delete_dir(path_)
-        status = utils.path_exists(path_)
+        utils.create_dir(self.path_)
+        utils.delete_dir(self.path_)
+        status = utils.path_exists(self.path_)
         self.assertFalse(status)
+        # for tearDown method.
+        utils.create_dir(self.path_)
```
paramsingh: This does not sound like a good practice, other tests will also need to create dir even if they don't need it. Let's fix the teardown function to only delete if the dir exists and do ko thing otherwise.

listenbrainz_spark/recommendations/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+
+TEST_PLAYCOUNTS_PATH = '/tests/playcounts.parquet'
+PLAYCOUNTS_COUNT = 100
+
+class TrainModelsTestClass(SparkTestCase):
```
paramsingh: TestCase for consistenct

listenbrainz_spark/utils.py
vansika: I have changed the setting of my editor long ago, It sometimes behaves weirdly :(
I will look into it :)

listenbrainz_spark/recommendations/tests/test_dataframe.py
vansika: Asserting the diff is not possible here IMO because after we go back x steps from to_date to get the from_date, we shift the from_date to the first of the month (let us say y steps). We know x but not y, and the diff can only be checked using from_date + x + y = to_date. 
I will try to find a better test though :)

listenbrainz_spark/recommendations/tests/test_dataframe.py
vansika: :)

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,155 @@
...
+from pyspark.sql import Row
+
+# path used in between test functions of this class
+LISTENS_PATH = '/test/listens/' # HDFS path to listens
+MAPPING_PATH = '/test/mapping.parquet'
+MAPPED_LISTENS_PATH = '/test/mapped_listens.parquet'
+
+class CreateDataframeTestCase(SparkTestCase):
+
+    date = datetime.utcnow()
```
paramsingh: I'd have preferred to put it as a member variable (self.date) instead of a class variable (cls.date) because it should be changed for every instance of the class but hey seems very nitpicky. Feel free to change it if you want.

listenbrainz_spark/recommendations/tests/test_dataframe.py
paramsingh: The current implementation looks good to me!

listenbrainz_spark/recommendations/tests/test_dataframe.py
vansika: I am really up for the suggestion but I think that __init__ of this class will override __init__ of unittest.TestCase. Is there any other way to put it as a member variable other than initializing inside every method?

paramsingh: Put it in setUp? You can define member variables anywhere, not just init.

listenbrainz_spark/recommendations/tests/test_dataframe.py
```diff
@@ -0,0 +1,154 @@
...
+LISTENS_PATH = '/test/listens/' # HDFS path to listens
+MAPPING_PATH = '/test/mapping.parquet'
+MAPPED_LISTENS_PATH = '/test/mapped_listens.parquet'
+
+class CreateDataframeTestCase(SparkTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.date = datetime.utcnow()
```
paramsingh: Not trying to be nitpicky but this still is a class variable and not a instance variable. At this point, I think we've spent too much time on this anyways, so let's leave it be but just pointing this out.

https://dev.to/ogwurujohnson/distinguishing-instance-variables-from-class-variables-in-python-81

vansika:
I wanted rob to review it before merging the PR. That was the whole purpose to create a separate branch to make it easy to review tests.

paramsingh:
Apologies, didn't know that. It can still be reviewed and changes made in a new PR if needed.


(also that sounds like the perfect thing to put in the action section of the PR description for the future)

vansika:
Yes, it can be, but I just like the notion of reviewing an open PR :p 

(Yeah. Will keep in mind)

listenbrainz_spark/recommendations/tests/test_dataframe.py
vansika: I know the difference between instance var and class var (in theory) , I am just not sure how to do a self.date in a class method. Maybe you can take over it and I will surely learn a lot from you on how to put this is in practice. 

But thank you for so many informative links :)

PR #76, opened 2020-01-05
Move import script to manage.py

paramsingh:
# Description

Move the import listens script to manage.py for consistency

# Problem

* JIRA ticket (_optional_): 

There was a different import script `import.py`. When importing the listens into the new setup, this confused me.

# Solution

Remove the import.py script and add an import command to manage.py


paramsingh:
@brainzbot retest this please

PR #80, opened 2020-01-15
Tests for ftp downloader and hdfs uploader

vansika:
# Description

Tests for 
- listenbrainz_spark/hdfs
- listenbrainz_spark/ftp


listenbrainz_spark/hdfs/upload.py
```diff
@@ -36,7 +37,7 @@ def process_json_listens(self, filename, data_dir, tmp_hdfs_path, schema):
                 tmp_HDFS_path (str): HDFS path where listens JSON has been uploaded.
         """
         start_time = time.time()
-        df = utils.read_json(tmp_hdfs_path, schema=schema).cache()
+        df = utils.read_json(tmp_hdfs_path, schema=schema)
```
paramsingh: why are we removing the cache function? I have some inkling that caching the file was needed for the import to work correctly because we delete the json later. Have you tested that the import works without this?

listenbrainz_spark/utils.py
```diff
@@ -264,3 +264,12 @@ def read_json(hdfs_path, schema):
     """
     df = listenbrainz_spark.session.read.json(config.HDFS_CLUSTER_URI + hdfs_path, schema=schema)
     return df
+
+def upload(hdfs_path, local_path):
```
paramsingh: since this is in utils, `upload_to_hdfs` would be a much clearer name. Earlier, it was called as `hdfs_connection.client.upload` which gave enough context, but `utils.upload` doesn't really help in understanding where it's getting uploaded.

listenbrainz_spark/ftp/download.py
```diff
@@ -50,6 +50,32 @@ def get_listens_file_name(self):
...
+                    If not provided, most recent dump will be downloaded.
+                ftp_dump_dir (str): FTP dir to find dump.
+                dump_id_pos (int): Unique identifier position in dump name.
+
+            Returns:
+                dest_path (str): Local path where dump has been downloaded.
+        """
+        self.connection.cwd(ftp_dump_dir)
+        dump = self.list_dir()
+        req_dump = self.get_req_dump(dump, dump_id, dump_id_pos)
```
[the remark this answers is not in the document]
vansika: req ~ required

listenbrainz_spark/hdfs/upload.py
vansika: >  Have you tested that the import works without this?

yes, for mappings (~ 6GB after decompression) , artist_relations (2GB after decompression) and listens (small set - incremental)

listenbrainz_spark/utils.py
vansika: thanks, will make the change :)

listenbrainz_spark/ftp/download.py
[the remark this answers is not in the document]
vansika: what can we name it? Do you want to suggest anything?

listenbrainz_spark/hdfs/tests/test_init.py
```diff
@@ -0,0 +1,100 @@
...
+
+                invalid_json_file_path = os.path.join(temp_dir, 'invalid.txt')
+                with open(invalid_json_file_path, 'w') as f:
+                    f.write('test file')
+                tar.add(invalid_json_file_path, arcname=os.path.join('temp_tar', 'invalid.txt'))
+            pxz.stdin.close()
+
+        return temp_archive
+
+    def test_upload_archive(self):
```
vansika: @paramsingh Can you look at this function? Sometimes it runs perfectly and sometimes throws the following error:-
`tarfile.ReadError: empty file`

When I opened this PR, it was running fine but now it fails suddenly. I tried thrice on my local machine, it worked for thw two times and failed for the third. Have you seen thisking of error before?

paramsingh: Sounds like a race condition where the file doesn't get completely written by the time the test starts trying to read it. There's two things you should try:

1. Make sure that the `pxz` objects are getting closed correctly like [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L257) and are being waited for like [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L259)
2. Add a sleep in the test after the write and before the read.

paramsingh:

